PR #8, opened 2021-03-26
Architecture/data layer

matuella:
Adds the data layer core pieces, such as `DatabaseRepository` and its implementation with `sembast`.

Unit tests on most of cases are also done.


lib/data/database_repository.dart
```diff
@@ -0,0 +1,157 @@
...
+        sink.add(transformedRecords);
+      },
+    );
+  }
+}
+
+extension StoreKeys on DatabaseStore {
+  @visibleForTesting
+  String get key {
+    switch (this) {
```
RaphaelOliveiraMoura, line 146: For this case is necessary a treatment for `default` of switch case?
Maybe add a custom throw error for this scenario, what do you think ?

matuella: This is not necessary because this project is null-safe, thus a switch case in a non-optional enum will always require all cases to be filled, and it will always match one of the enumerator cases

RaphaelOliveiraMoura: Make sense, tks =)

RaphaelOliveiraMoura (approved):

ggirotto (changes requested):
Although the `DatabaseRepositoryImpl` receives a database instance on its constructor, this data layer is not ready to handle multiple database instances. Have you though about this multiple databases necessity? How would you manage this in the current structure?

lib/data/database_repository.dart
```diff
@@ -0,0 +1,157 @@
...
+    required DatabaseStore store,
+  }) async {
+    final storeMap = stringMapStoreFactory.store(store.key);
+
+    final transformer = _snapshotSerializerTransformer(serializer);
+    return storeMap.query().onSnapshots(_db).transform(transformer);
+  }
+
+  /// Transforms a list of `sembast` snapshot records into a list of objects parsed by [serializer]
+  StreamTransformer<List<RecordSnapshot<String, Map<String, Object?>>>, List<T>>
```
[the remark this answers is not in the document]
matuella: There is no way to create custom types that aren't functions. This [might be possible in 2.13](https://github.com/dart-lang/language/issues/65#issuecomment-808399347), but not right now

matuella:
@ggirotto 
> Although the `DatabaseRepositoryImpl` receives a database instance on its constructor, this data layer is not ready to handle multiple database instances. Have you though about this multiple databases necessity? How would you manage this in the current structure?

The idea is that we should not need to instantiate multiple databases just for the sake of performance. If there is a problem with performance, even if they are mitigated when split into multiple files, those problems will inevitably occur even if a single DB has too many _elements_. So if this is a problem, there are plausible workarounds, like parameterizing the constructor/openDB function. Although the underlying problem (which is the library not being performant with a heavy load of elements) is what should be solved instead, not creating multiple DB instances.

With this in mind, I have my doubts that it will be really necessary in the future, as we should eventually **need** to go to cloud to provide such amount of data, and not storing thousands of MBs for decks that the user will never use (possibly the majority) on the device.

ggirotto (approved):

PR #9, opened 2021-03-29
Architecture/domain layer

matuella:
Adds the domain layer core pieces: all initial models, serializers and its tests.


lib/domain/serializers/card_block_serializer.dart
```diff
@@ -0,0 +1,44 @@
...
+  @override
+  Map<String, dynamic> mapOf(CardBlock block) => <String, dynamic>{
+        'type': block.type.raw,
+        'rawContents': block.rawContents,
+      };
+
+  CardBlockType _typeFromRaw(String raw) => CardBlockType.values.firstWhere(
+        (type) => type.raw == raw,
+        orElse: () {
+          throw SerializationError("Failed to find a CardBlockType with the raw vale of '$raw'");
```
ggirotto, line 26: value of

lib/domain/models/card_execution.dart
```diff
@@ -0,0 +1,47 @@
...
+    required this.started,
+    required this.finished,
+    required this.question,
+    required this.answer,
+    required this.answeredDifficulty,
+  })   : assert(started.isBefore(finished)),
+        assert(question.isNotEmpty),
+        assert(answer.isNotEmpty);
+
+  final DateTime started;
```
ggirotto, line 20: IMO `startedAt` and `finishedAt` are better. Just `started` and `finished` may confuse with a boolean like `isFinished`

lib/domain/models/resource.dart
```diff
@@ -0,0 +1,22 @@
+import 'package:memo/data/database_repository.dart';
+import 'package:meta/meta.dart';
+
+@immutable
+class Resource extends KeyStorable {
+  Resource({required String id, required this.description, required this.tags, required this.url})
+      : assert(tags.isNotEmpty, 'tags must have at least one element'),
```
ggirotto, line 7: I think that today we don't have any case of a Resource with empty URL, so we could assert it here as well

ggirotto: Also, we must differ the resources types so we can customize them in the UI like `video`, `article`, `book`, etc..

ggirotto (changes requested):

lib/domain/models/card_execution.dart
matuella: I think `startedAt` comes from a Swift background where function signatures can have the full argument like `myFunc(startedAt startedDate:)`. Just like we don't name an `int index` as `int indexAt`. This is a personal opinion and while I agree, I think it should be `startDate` instead of suffixing with `at`

lib/domain/models/resource.dart
matuella: I don't think these assertions should be this fine-grained, because this opens to discussion: well, it should also be a valid URL, and it should always be HTTPS, and the list goes on. So I don't think we should validate strings at all, other than they should be null-safe required.

lib/domain/models/resource.dart
```diff
@@ -14,6 +14,11 @@ class Resource extends KeyStorable {
   /// This is useful in cases where we must match a `Deck.tags` with each available resources
   final List<String> tags;
 
+  /// Metadata that describes which type of [url] this resource refers to
+  ///
+  /// Example types: "video", "article", etctera.
+  final String type;
```
ggirotto, line 20: Why using a String instead a enum? I mean, we will have to customize this Resource in the UI, so we will need to compare these types

ggirotto (changes requested):

lib/domain/models/resource.dart
matuella: No need to force an enumerator here. The rationale is that we can have a default fallback case if none is match (like an default icon/ui customization). If you require an enumerator here, we may have to crash at runtime when reading such resources (or even make this type optional). So, in my opinion it's better to iterate over a loose type instead of the enumerator in this particular case.

ggirotto: @matuella  We can merge both solutions: an enum with a "unknown" type. This way we doesn't need to crash at runtime nor compare strings constants to create a custom UI. My points against using a `String` here are 1. We will have silent type inconsistencies (we will show a default UI behavior, but someone see's it, we will neve know that a Resource is with the wrong type) and 2. We will have to compare string constants somewhere to customize the UI, like `if (type == "book") { ... }`

matuella: I think this is more of a personal opinion than what we should seek here, but I guess the work is the same with enumerators and strings

ggirotto (approved):

Issue #16, opened 2021-03-31
Setup GithubAction with Flutter


gustavohd18:
**Caso de Uso/Use Case:**
If we have GitActions for Flutter set, it would be possible to check Lint, tests, etc in each PR.

**Proposta/Proposal:**
Creation of Gitactions for Flutter.

matuella:
I would also add that when we are uploading to AppStore, ideally it should also be done by the CI (through `fastlane`), and not manually.

Thanks for the suggestion. Even though we aren't planning to pick this improvement for now, it will be a nice enhancement.

ggirotto:
It would be an awesome enhancement to the project. If you have experience with Github Actions and want to contribute with the project, I suggest to open a PR with this feature, would be great to have this integration in the repo.

matuella:
Closed with #18

PR #18, opened 2021-03-31
Add github actions workflow

RaphaelOliveiraMoura:
Setup of CI with github actions 

**Proposed steps**:
- Installation of flutter and flutter dependences
- Execution of flutter formatter, analyser and linter
- Execution of tests

**Trigger**: on pull requests

For the future will be interesting add the CD setup with `fastlane` as the mentioned by @matuella on issue #16.
Some details would be important set here too, like flutter version or when the process will be trigged. 

This setup was inspired on this [gist](https://gist.github.com/aagarwal1012/eb2551683d81cf47d60868ab607520b1)

This changes have been discussed on issue #16 

## Pre-launch Checklist

- [x] Eu li o [CONTRIBUTING.md](../CONTRIBUTING.md) e segui as sugestões descritas ali sobre boas práticas de código / I
read [CONTRIBUTING.md](../CONTRIBUTING.md) and followed the suggestions proposed there about writing good code;
- [x] Eu listei pelo menos um issue relacionado à este PR / I listed at least one issue related to this PR;
- [ ] Eu adicionei novos testes para cobrir as mudanças que eu fiz neste PR, ou este PR é uma exceção / I added new 
tests that cover the changes made in this PR or this PR is a test-exception;
- [ ] Eu atualizei o [CHANGELOG.md](./../CHANGELOG.md) com as mudanças feitas neste PR / I updated 
[CHANGELOG.md](./../CHANGELOG.md) with the changes made in this PR;
- [ ] Eu atualizei/adicionei comentários relevantes nas modificações efetuadas / I updated or added relevant docs in the
modifications I've made;
- [x] Todos os testes previamente existentes e os novos criados por mim estão finalizando com sucesso / All the existing
tests - the old and the new ones - are finishing with success.

### DIFF
--- a/.github/workflows/flutter-ci.yml
+++ b/.github/workflows/flutter-ci.yml
@@ -18,5 +18,7 @@ jobs:
       - run: flutter pub get
       - run: flutter analyze .
       - run: flutter test
-      - run: flutter build appbundle
-      - run: flutter build ios --release --no-codesign
+      - run: flutter build appbundle --dart-define=ENV=${{env.FLUTTER_ENV}}
+      - run: flutter build ios --release --no-codesign --dart-define=ENV=${{env.FLUTTER_ENV}}
+        env:
+          FLUTTER_ENV: ${{ secrets.FLUTTER_ENV || 'PROD' }}
--- a/.github/workflows/flutter-ci.yml
+++ b/.github/workflows/flutter-ci.yml
@@ -19,3 +19,4 @@ jobs:
       - run: flutter format --set-exit-if-changed .
       - run: flutter analyze .
       - run: flutter test
+      - run: flutter build appbundle
--- a/.github/workflows/flutter-ci.yml
+++ b/.github/workflows/flutter-ci.yml
@@ -31,9 +31,9 @@ jobs:
       - name: Building for android
         run: flutter build appbundle --dart-define=ENV=${{env.FLUTTER_ENV}}
         env:
-          FLUTTER_ENV: ${{ secrets.FLUTTER_ENV || 'PROD' }}
+          FLUTTER_ENV: ${{ github.event.pull_request.base.ref == 'main' && 'PROD' || 'DEV' }}
 
       - name: Building for ios
         run: flutter build ios --release --no-codesign --dart-define=ENV=${{env.FLUTTER_ENV}}
         env:
-          FLUTTER_ENV: ${{ secrets.FLUTTER_ENV || 'PROD' }}
+          FLUTTER_ENV: ${{ github.event.pull_request.base.ref == 'main' && 'PROD' || 'DEV' }}
--- a/.github/workflows/flutter-ci.yml
+++ b/.github/workflows/flutter-ci.yml
@@ -16,7 +16,6 @@ jobs:
           channel: 'stable'
 
       - run: flutter pub get
-      - run: flutter format --set-exit-if-changed .
       - run: flutter analyze .
       - run: flutter test
       - run: flutter build appbundle
--- a/.github/workflows/flutter-ci.yml
+++ b/.github/workflows/flutter-ci.yml
@@ -1,0 +1,21 @@
+name: Flutter CI
+
+on: pull_request
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-java@v1
+        with:
+          java-version: '12.x'
+
+      - uses: subosito/flutter-action@v1
+        with:
+          channel: 'stable'
+
+      - run: flutter pub get
+      - run: flutter format --set-exit-if-changed .
+      - run: flutter analyze .
+      - run: flutter test
--- a/.github/workflows/flutter-ci.yml
+++ b/.github/workflows/flutter-ci.yml
@@ -6,19 +6,34 @@ jobs:
   build:
     runs-on: macos-latest
     steps:
-      - uses: actions/checkout@v1
-      - uses: actions/setup-java@v1
+      - name: Checkout repository
+        uses: actions/checkout@v1
+
+      - name: Setup Java
+        uses: actions/setup-java@v1
         with:
           java-version: '12.x'
 
-      - uses: subosito/flutter-action@v1
+      - name: Setup Flutter
+        uses: subosito/flutter-action@v1
         with:
           channel: 'stable'
 
-      - run: flutter pub get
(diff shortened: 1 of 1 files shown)


gustavohd18 (changes requested):
I think you should add `flutter build` too.

RaphaelOliveiraMoura:
> I think you should add `flutter build` too.

even now that doesn't have a Continuous Integration with the application stores ? Because, what will be the purpose of the build script for this case ?

gustavohd18:
> > I think you should add `flutter build` too.
> 
> even now that doesn't have a Continuous Integration with the application stores ? Because, what will be the purpose of the build script for this case ?

I think with a build script, we could detect cases like writing a wrong variable or something like that. Only with the build test or Lint, they are not possible to catch. But is just suggested.

RaphaelOliveiraMoura:
> > > I think you should add `flutter build` too.
> > 
> > 
> > even now that doesn't have a Continuous Integration with the application stores ? Because, what will be the purpose of the build script for this case ?
> 
> I think with a build script, we could detect cases like writing a wrong variable or something like that. Only with the build test or Lint, they are not possible to catch. But is just suggested.

Make sense. Very tks 😃

gustavohd18:
Cool!

ggirotto (changes requested):
Hey @RaphaelOliveiraMoura, thanks for contributing with the project!

I agree with you in the point that we don’t have a CI/CD to deploy the project to the stores yet, but I think we can create an initial version of it from your PR. I understand as an initial version a script that builds and run the tests of the project, similar as you did. We can latter evolve to something more sophisticated, but lets start concise. 

I’m not very experienced with Github Actions, but I can see some points of failure in this script:

- It just builds the Android target of the app, but we deploy to iOS as well;
- As consequence of the previous point, your script is also running on an ubuntu machine, which cannot build the iOS target as far is I know;
- I don’t see the necessity of formatting the code by using flutter format, since the project lint should handle it properly;
- As described in the [ARCHITECTUREmd environment section](https://github.com/olmps/memo/blob/main/ARCHITECTURE.md#environment), your build must specify the build environment, so your flutter build appbundle will probably become something like flutter build appbundle —dart-define=ENV=PROD (same to iOS target: flutter build ios —release -dart-define=ENV=PROD

Having a script that can build and run tests seems as a good v1 to me, but the above issues needs to be fixed. 

Thanks for contributing!

lizzi2122 (approved):

lizzi2122 (approved):
Ok

gustavohd18 (approved):

lizzi2122 (approved):

lizzi2122 (approved):

RaphaelOliveiraMoura:
Thanks @ggirotto for the feedback,

I think every points you mentioned was covered:
- was removed the format code script
- was changed machine that ci run for mac-os machine to build app at both: android and ios
- the last point I was with some doubts if is correct, but I put the possibility to setup a variable called `ENV` passing to the dart-defines, this env var has a default value: `PROD`, and can be change on github secrets at repository, this is correct ?

ggirotto:
@RaphaelOliveiraMoura Thanks a lot for your effort, we are getting close to the final version. About your last point, the env being stored in the Github Actions Secret, I think it shouldn't be there. What I suggest, is to use environments as follows:
- Pull Requests that are pointing to `main` branch, use `PROD` environment
- Pull Requests that are pointing to any other branch that is not main, use `DEV` environment

I don't think that the env should be stored in secrets, since it's not a secret at all.
Here's an example of what I'm saying: https://stackoverflow.com/questions/63817052/github-actions-run-step-only-for-certain-pull-request-base-branches

RaphaelOliveiraMoura:
@ggirotto done, PRs created to main branch will be run with ENV=PROD and PRs to other branches will run with ENV=DEV

Was removed the possibility to change the environment at secrets. And I dont used the if statement for each scenario, instead I preferred to use a ternary operator to populate the env variable with PROD or DEV based on pull request base reference. I think this way was more clearly and don't need duplications for PROD and DEV scenarios.

ThiagoBfim:
Hello, it would be nice to include coverage as well.

Here's an example: https://admcpr.com/ci-for-flutter-with-github-actions-part-2-code-coverage-and-matrix/

.github/workflows/flutter-ci.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      - name: Setup Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: '12.x'
+
+      - name: Setup Flutter
+        uses: subosito/flutter-action@v1
+        with:
+          channel: 'stable'
```
ggirotto, line 20: I found that you can replace these both setup actions by using Google's Dart container:

```
container:
      image: google/dart:2.12
```

References:
https://github.com/felangel/bloc/blob/master/.github/workflows/bloc.yaml
https://www.bradcypert.com/dart-packages-with-github-actions/

ggirotto (changes requested):
@RaphaelOliveiraMoura Sorry the delay and thanks for the update. I commit to be faster on answering the PRs so we can speed up these modifications. Also, I'm including a new suggestion that would minimize the amount of time that the CI runs and cut-off the setup steps of the script.

ggirotto:
@ThiagoBfim Good suggestion, but I think it's easier for me to do that in another PR, since it involves creating the project account in CodeCov and adding the secrets to the Actions. I'll create an issue to address this enhancement

RaphaelOliveiraMoura:
@ggirotto I think macos doesnt have support to docker, so its not possible add the container to pipeline.
https://github.community/t/why-is-docker-not-installed-on-macos/17017

I see two paths:
- Use ubuntu machine with pre setup with container and build just for android (will reduce the ci time)
- Use macos without pre setup container with docker and build for both: android and ios (will not reduce the ci time)

What do you think about that?

ggirotto:
@RaphaelOliveiraMoura That's sad :/. 

Lets go back to setup without the container and building for iOS. I believe we should keep the iOS build because we have native code written in the future, the CI must be capable to ensure that it's not breaking

RaphaelOliveiraMoura:
@ggirotto done, I will continue searching about the topic to try reduce the pipeline time, if I found something relevant I will open a new issue or discussion.
Very thanks for engagement.

ggirotto:
@RaphaelOliveiraMoura We say the same, thanks for contributing with the project!

ggirotto (approved):

PR #34, opened 2021-04-10
Adds TextTag, PrimaryTextTag and SecondaryTextTag widgets

matuella:


lib/application/widgets/text_tag.dart
```diff
@@ -0,0 +1,33 @@
...
+
+/// Creates a new [Text] that resembles a tag
+///
+/// It's important to no confuse this tag-styled component with the material's `Chip`, which is used in more complex
+/// - usually interactable - scenarios.
+class TextTag extends StatelessWidget {
+  const TextTag(
+    this.text, {
+    required this.backgroundColor,
+    required this.padding,
```
ggirotto, line 12: I couldn't understand the use-case that needs this `padding` to be parametrized in the Widget

lib/application/widgets/theme/themed_text_tag.dart
```diff
@@ -0,0 +1,41 @@
+import 'package:flutter/material.dart';
```
ggirotto, line 1: In this file as an example, both implementations use the same padding: `context.allInsets(Spacing.xxSmall)`

ggirotto (changes requested):

lib/application/widgets/theme/themed_text_tag.dart
matuella: Well, it splits the responsibility of the implementation versus what the component needs.

My perspective is that if you can split things into smaller workable and independent stuff, without adding too much overhead to the code, you should do that. In this case, for instance, I don't need to force it down through the `TextTag`, as the implementers that are already exposed to a lot of distinct frameworks (like flutter hooks, riverpods, layoutr), can remain being "dirty", while the core can remain mostly pristine, only accessing fundamental frameworks, like material.

That's one of the reasons why I have split the `widgets/` folder. It's described a little bit more in ARCHITECTURE.md

ggirotto: Got it

ggirotto (approved):

PR #35, opened 2021-04-10
Adds AssetIconButton material widget

matuella:


lib/application/widgets/material/asset_icon_button.dart
```diff
@@ -0,0 +1,27 @@
...
+  final void Function()? onPressed;
+
+  @override
+  Widget build(BuildContext context) {
+    final imageAsset = AssetImage(asset);
+    final icon = ImageIcon(imageAsset, color: iconColor);
+
+    // Even though `ImageIcon` already uses the theme to specify the icon size, we must enforce this through the
+    // `IconButton` instance, otherwise it will be overridden by the default argument
+    final normalizedIconSize = iconSize ?? Theme.of(context).iconTheme.size;
```
ggirotto, line 19: Why `iconTheme.size` can be null? Can we have a custom theme that doesn't specify the icons default size?

ggirotto (changes requested):

lib/application/widgets/material/asset_icon_button.dart
matuella: Because this is not a required argument in the material's `ThemeData`, so it's plausible for this to not be available if it stops existing

ggirotto (dismissed):

PR #42, opened 2021-04-12
Overall tests enhancements

matuella:


test/domain/models/card_block_test.dart
```diff
@@ -8,7 +8,7 @@ void main() {
       () {
         CardBlock(type: CardBlockType.text, rawContents: '');
       },
-      throwsA(isA<AssertionError>()),
```
ggirotto, line 11: Nice improvement

test/data/database_repository_test.dart
```diff
@@ -97,14 +97,18 @@ void main() {
     expect(objects.isEmpty, true);
   });
 
-  test('DatabaseRepositoryImpl should emits objects with success', () async {
-    final listFakeObjects = List.generate(2, (_) => fakeObject);
-
+  test('DatabaseRepositoryImpl should emit new events when listening to store updates', () async {
```
ggirotto, line 100: Following the same enhancements you made in the other files, whynot just `should emit new events when listening to store updates`?

ggirotto (approved):

test/data/database_repository_test.dart
matuella: That's because not using the tested file's name is only useful when we are describing a group. The updates that I made in this PR are those that were duplicated ([commit](https://github.com/olmps/memo/pull/42/commits/397716312d66012b09413a421afd13ed9c5b8b11))

PR #44, opened 2021-04-16
Adds Collections Page

matuella:


test/application/widgets/theme/themed_tab_bar_test.dart
```diff
@@ -0,0 +1,42 @@
...
+              tabs: const [Text('tab 1'), Text('tab 2')],
+            );
+          },
+        ),
+      ),
+    );
+
+    final bottomDividerFinder = find.byType(Container).first;
+    final size = tester.getSize(bottomDividerFinder);
+    final bottomLeftPosition = tester.getBottomRight(bottomDividerFinder);
```
ggirotto, line 29:
```suggestion
    final bottomLeftPosition = tester.getBottomLeft(bottomDividerFinder);
```

test/application/widgets/theme/themed_tab_bar_test.dart
```diff
@@ -0,0 +1,42 @@
...
+    final bottomRightPosition = tester.getBottomRight(bottomDividerFinder);
+
+    final parentFinder = find.byType(ThemedTabBar).first;
+    final bottomLeftParentPosition = tester.getBottomRight(parentFinder);
```
ggirotto, line 33:
```suggestion
    final bottomLeftParentPosition = tester.getBottomLeft(parentFinder);
```

lib/application/widgets/theme/themed_tab_bar.dart
```diff
@@ -0,0 +1,36 @@
...
+  Widget build(BuildContext context) {
+    final memoTheme = useTheme();
+    final selectedColor = memoTheme.secondarySwatch.shade400;
+    final unselectedColor = memoTheme.neutralSwatch.shade700;
+    const borderHeight = dimens.tabBarBorderHeight;
+
+    final tabBar = TabBar(
+      controller: controller,
+      tabs: tabs,
+      indicatorWeight: borderHeight * 2,
```
ggirotto, line 23: Is this condition always true? I mean, it seems like an arbitrary multiplication. If `dimens.tabBarBorderHeight` changes for some reason, this expression keeps valid?

lib/application/pages/home/collections/collections_list_view.dart
```diff
@@ -0,0 +1,235 @@
+import 'dart:ui' as ui;
+
+import 'package:flutter/material.dart';
+import 'package:flutter_hooks/flutter_hooks.dart';
+import 'package:hooks_riverpod/hooks_riverpod.dart';
+import 'package:layoutr/common_layout.dart';
+import 'package:memo/application/constants/colors.dart' as colors;
```
ggirotto, line 7: Already mentioned below about the reexport. If it's possible, it would be a great improvements in the import list here

lib/application/view-models/home/collections_vm.dart
```diff
@@ -0,0 +1,173 @@
...
+class CollectionsVMImpl extends CollectionsVM {
+  CollectionsVMImpl() : super(LoadingCollectionsState(availableFilters.first)) {
+    _loadCollections();
+  }
+
+  @override
+  Future<void> updateCollectionsFilter(CollectionsFilter filter) async {
+    state = LoadingCollectionsState(filter);
+
+    // TODO(matuella): attach logic
```
ggirotto, line 69: Is this a missing TODO or it will be updated in a future task? Could be an issue

lib/application/pages/home/collections/collections_list_view.dart
```diff
@@ -0,0 +1,235 @@
...
+    return ListView.builder(
+      itemCount: items.length,
+      itemBuilder: (context, index) {
+        // Builds the respective widget based on the item's `CollectionItemMetadata` subtype
+        final item = items[index];
+
+        if (item is CollectionsCategoryMetadata) {
+          return _CollectionsSectionHeader(
+            title: item.name,
+            onTap: () {},
```
ggirotto, line 39: You added TODOs in `CollectionsVM` missing implementations, shouldn't have a TODO here as well?

lib/application/pages/home/collections/collections_list_view.dart
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    final loadedState = state as LoadedCollectionsState;
+    final items = loadedState.collectionItems;
+
+    return ListView.builder(
+      itemCount: items.length,
+      itemBuilder: (context, index) {
+        // Builds the respective widget based on the item's `CollectionItemMetadata` subtype
+        final item = items[index];
```
ggirotto, line 34: I liked the way how you built `loadedState.collectionItems` adding the headers in the same array as the items so the list can build them in the correct order

lib/application/pages/home/collections/collections_page.dart
```diff
@@ -0,0 +1,57 @@
...
+    );
+    final tabs = availableFilters.map(_widgetForTab).toList();
+
+    // Adds listener once (in the first build call)
+    useEffect(() {
+      collectionsTabController.addListener(() {
+        final currentState = context.read(collectionsVM.state);
+
+        // We want to update only when the indexIsChanging (because this listener is called multiple times by the tab
+        // controller)  and if the current tab index is different from the index of the current filter, so they are
```
ggirotto, line 26: Double space

ggirotto: Isn't the lint complaining about it?

lib/application/view-models/home/collections_vm.dart
```diff
@@ -0,0 +1,173 @@
...
+final collectionsVM = StateNotifierProvider<CollectionsVM>((_) => CollectionsVMImpl());
+
+enum CollectionsFilter { explore, review }
+const availableFilters = CollectionsFilter.values;
+
+abstract class CollectionsVM extends StateNotifier<CollectionsState> {
+  CollectionsVM(CollectionsState state) : super(state);
+
+  /// Updates the current [state] with [filter]
+  Future<void> updateCollectionsFilter(CollectionsFilter filter);
```
ggirotto, line 15: I believe that the `filter` word here is not a good decision. If we add inner filters in `Explorar` and `Revisar` tabs to filter the collections from each tab, these filters naming would conflict heavily with this filter definition. Unless all filters (Explorar & Revisar + inner filters) could be handled as the same thing - and I don't know if it's possible with today's implementation - it would have no problem, but I suspect that this may not be possible with the filter structure.

ggirotto (changes requested):

ggirotto:
@matuella Also, I noticed that you are pointing directly to `main` branch. What do you think about what we discussed some days ago about not using the `development` branch and work only with `main` + release tags?

lib/application/constants/strings.dart
```diff
@@ -0,0 +1,16 @@
+// This is a temporary constants file related to all strings
+//
+// TODO(matuella): update this approach to one that takes the localization (BuildContext)
```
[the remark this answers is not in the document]
matuella: But that's not a issue, as we may never update the Localization, thus this won't need to never use `BuildContext`

lib/application/pages/home/home_page.dart
```diff
@@ -1,9 +1,13 @@
 import 'package:flutter/material.dart';
 import 'package:flutter_hooks/flutter_hooks.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:memo/application/constants/images.dart' as images;
```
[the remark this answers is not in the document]
matuella: I don't like re-exporting here in dart, I think this is a very JS way to deal with things and how dart works importing stuff, I think in the overall perspective, just shadow stuff when re-exporting everything through a single point. Also, there is a really high chance that someone, somewhere, will import the individual file and not the entry index file for that particular folder, which will make the imports inconsistent

lib/application/widgets/theme/themed_tab_bar.dart
matuella: yes because this is double the back indicator height

lib/application/view-models/home/collections_vm.dart
matuella: future task

lib/application/pages/home/collections/collections_list_view.dart
matuella: no, this is a particular individual issue that we haven't tracked in our tasks as usual, I've already create the pending tasks for this

matuella:
> @matuella Also, I noticed that you are pointing directly to `main` branch. What do you think about what we discussed some days ago about not using the `development` branch and work only with `main` + release tags?

That is exactly what I am doing, as of today, the `development` serves no clear purpose other than being an intermediate branch that benefits nothing. If people want to use the `master` branch instead of the releases, they know that they are the latest changes, meaning that while it should not break, it should also not be directly used for production development, meaning that people should use the released versions

lib/application/view-models/home/collections_vm.dart
matuella: What is your suggestion on this naming? Because that's what they are to the View Model, a filter. If the UI uses them as TABS, that's an UI decision, and not the VM's

lib/application/pages/home/collections/collections_list_view.dart
matuella: Already added my 2 cents on why re-exporting in dart is not a good overall decision IMHO.

When IDEs (vscode) and other lint decisions are improved, we may want to discuss this again, but as things are today, there is more to lose than to gain if following a re-export entry file

lib/application/constants/strings.dart
```diff
@@ -0,0 +1,16 @@
...
+//
+// Even if the v1 for memo will only be available in ptBR, splitting random strings throughout the code will make it
+// hard when we decide to support multipla locales.
```
RaphaelOliveiraMoura, line 6: Typo: hard when we decide to support **multipla** locales.

lib/application/constants/strings.dart
```diff
@@ -0,0 +1,16 @@
...
+
+const collectionsNavigationTab = 'Coleções';
+const progressNavigationTab = 'Progresso';
+
+const collectionsExploreTab = 'Explorar';
+const collectionsReviewTab = 'Revisar';
+
+const collectionsSectionHeaderSeeAll = 'VER TODOS';
```
RaphaelOliveiraMoura, line 14: Wouldn't it be better to keep the responsibility of using this text in upper case for the widget that uses it?

lib/application/constants/strings.dart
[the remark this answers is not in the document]
ggirotto: But if this feature may never be implemented, what is the purpose of this TODO? I mean, or this is an enhancement issue or this TODO shouldn't be here because if someday we decide to implement it, it should be handled as a new feature.

Otherwise this TODO will stay here indefinitely, because as you said, this may never be updated

lib/application/view-models/home/collections_vm.dart
ggirotto: What about `section`? It indicates that they have a separation but it's not UI specific as `tab` is

lib/application/constants/strings.dart
matuella: Agreed!

RaphaelOliveiraMoura (approved):

lib/application/view-models/home/collections_vm.dart
ggirotto: I can't understand why you see `section` as "UI specific" since it doesn't leak how the UI is structured, just that it has a "major" division. But my point here is not about the `section` thing, I could suggest a list of other synonyms that are worst in my opinion but fit here, like `segment`, `module`, etc... We eventually would find one that it's not UI specific.

But the main point here is about the conflict that may arise if this screen receives a categories filter for each tab, and I don't know if you agree or disagree about it. If you think that this could be handled with two different filter functions like `updateCollectionsFilter` and `updateCategoriesFilter`, we can keep the filter naming, my concern is about it being 
 confuse, that's why I suggest the naming change. But again, you haven't expressed yourself about this possible conflict.

matuella:
@ggirotto

ggirotto (approved):

PR #45, opened 2021-04-20
Adds `ProgressPage`

matuella:


lib/application/constants/strings.dart
```diff
@@ -16,3 +16,30 @@ const collectionsReviewTab = 'Revisar';
 const collectionsSectionHeaderSeeAll = 'Ver todos';
 
 const collectionsMemoryStability = 'Estabilidade da Memória';
+
+const progressTotalStudyTime = 'HORAS TOTAIS DE ESTUDOS';
```
ggirotto, line 20: Following the same suggestion from the last PR, what about using `toUppercase` in this string use-case?

https://github.com/olmps/memo/pull/44#discussion_r615843902

lib/application/pages/home/progress/progress_page.dart
```diff
@@ -0,0 +1,222 @@
...
+        progressValue: loadedState.easyMemosPercentage,
+        centerLabel: strings.squintingFaceWithTongue,
+        semanticLabel: strings.progressEasyMemosIndicatorLabel,
+      ),
+      title: _buildAlternateStyleTextBox(
+        context,
+        texts: [loadedState.readableEasyMemosPercentage, strings.percentSymbol],
+      ),
+      description: strings.progressTotalEasyMemos,
+    );
```
ggirotto, line 80: All this progress components are very similar. Have you tried to create an abstract function to create them? Or the amount of parameters from this function makes the abstraction approach worst than the duplicate code?

matuella: But nothing is duplicate here, they are simply passing arguments to a widget and two functions that build two other dependent widgets. If one would create a function, it would possibly take the same amount of "lines", as the arguments from each other are completely different (other than using `context` twice and `strings.percentSymbol` being shared among them all)

lib/application/constants/strings.dart
ggirotto: @matuella What about this one? Do you think that it makes sense?

ggirotto (approved):

PR #48, opened 2021-04-26
Services related to the Execution functionality

matuella:


lib/domain/services/collection_services.dart
```diff
@@ -1,15 +1,69 @@
...
 
+/// Handles all domain-specific operations pertaining to one or multiple [Collection]
 abstract class CollectionServices {
+  /// Retrieves all available [CollectionStatus] and keeps listening to any changes made to them
+  Future<List<Stream<CollectionStatus>>> listenToAllCollectionsStatus();
+
+  /// Retrieves all available [Collection]
   Future<List<Collection>> getAllCollections();
+
+  /// Retrieves the [Collection] (with [collectionId]) current memory stability
```
ggirotto, line 15: I would like to suggest a doc pattern that we could follow: we could refer to the id properties as it were their own entities

Example:
`Retrieves the [collectionId] current memory stability`

I noticed that in most of the situations, the id property has the name of its entity, so we could refer to it as it was the entity itself

lib/domain/services/collection_services.dart
```diff
@@ -1,15 +1,69 @@
...
+
+    var averageStability = 0.0;
+    allCollectionMemos.forEach((memo) {
+      final stability = memoryServices.evaluateMemoryStability(memo);
+      if (stability != null) {
+        averageStability += stability;
+      }
+    });
+
+    return averageStability / allCollectionMemos.length;
```
ggirotto, line 67: I think that the discussion that we had some days ago in Discord about counting the `pristine` memos in the average math could be summarized here

lib/domain/services/execution_services.dart
```diff
@@ -0,0 +1,163 @@
...
+import 'package:memo/data/repositories/user_repository.dart';
+import 'package:memo/domain/enums/memo_difficulty.dart';
+import 'package:memo/domain/isolated_services/memory_stability_services.dart';
+import 'package:memo/domain/models/collection.dart';
+import 'package:memo/domain/models/memo.dart';
+import 'package:memo/domain/models/memo_execution.dart';
+
+/// Handles all domain-specific operations pertaining to the act of executing a [Collection] and its [Memo]s
+abstract class ExecutionServices {
+  /// Retrieves a suitable chunk of [Memo]s when executing a [Collection] of [collectionId]
```
ggirotto, line 13: Is this function used DURING an execution? If it is, disregard the following suggestion.

IMO this could mention that the this chunk can be used - in the future - in an execution flow instead using `when executing`. Using `when executing` suggests to me that this chunk can be used during an execution instead before it.

I suggest something like:
`Retrieves a suitable chunk of [Memo]s of [collectionId] to be executed`

lib/domain/services/execution_services.dart
```diff
@@ -0,0 +1,163 @@
...
+  final MemoRepository memoRepo;
+  final MemoExecutionRepository executionsRepo;
+  final CollectionRepository collectionRepo;
+
+  final MemoryStabilityServices memoryServices;
+
+  @override
+  Future<List<Memo>> getNextExecutableMemosChunk({required String collectionId}) async {
+    final user = await userRepo.getUser();
+    // This is the size of our chunk
```
ggirotto, line 50: IMO naming as `final chunkSize` would dispense this comment

ggirotto: The developer can infer that is the daily goal due to `user.dailyMemosGoal`

lib/domain/services/execution_services.dart
```diff
@@ -0,0 +1,163 @@
...
+    final executedMemoPerStability =
+        executedMemos.asMap().map((key, value) => MapEntry(memoryServices.evaluateMemoryStability(value)!, value));
+    final sortedStabilityKeys = executedMemoPerStability.keys.toList()
+      ..sort((stabilityA, stabilityB) => stabilityA.compareTo(stabilityB));
+
+    final executableMemosChunk = <Memo>[];
+
+    // Keep adding the remaining pristine memos and then the sorted-by-memory-stability memos until the chunk has met
+    // its required size, or if there ar no more memos to be added
+    while (dailyGoal > executableMemosChunk.length || (pristineMemos.isEmpty && executedMemoPerStability.isEmpty)) {
```
ggirotto, line 80: Why not initialize `executableMemosChunk` with all memos from `pristineMemos` (since they are not greater than the daily goal - and then adds the first Y memos from `executedMemoPerStability` where Y is `dailyGoal - pristineMemos.length`?

ggirotto:
```dart
final executableMemosChunk = pristineMemos;
final remainingMemos = dailyGoal - pristineMemos.length;
while (dailyGoal > executableMemosChunk.length) {
    executableMemosChunk.add(executedMemoPerStability[sortedStabilityKeys]!);
    executedMemoPerStability.remove(sortedStabilityKeys);
}
```

lib/domain/services/execution_services.dart
```diff
@@ -0,0 +1,163 @@
...
+    // Uses the `executionsAmounts` to get the updated `executionsAmount` for the user and the associated collection
+    executionsAmounts.forEach((key, value) {
+      final userCurrentValue = userUpdatedExecutions[key] ?? 0;
+      userUpdatedExecutions[key] = userCurrentValue + value;
+
+      final collectionCurrentValue = collectionUpdatedExecutions[key] ?? 0;
+      collectionUpdatedExecutions[key] = collectionCurrentValue + value;
+    });
+
+    // Dispatch all updates simultaneously
```
ggirotto, line 147: What happens if one fail? Shouldn't it be made using a transaction?

ggirotto (changes requested):

lib/domain/services/execution_services.dart
matuella: Not during, just before executing.
The execution doesn't change anything when it's executing, that's a decision that I've made to make things way easier and should be implied when the user decides to leave an execution session, as a form of "discarding what he is doing at the moment", even though he may have executed one or more memos while at it.

lib/domain/services/collection_services.dart
matuella: The problem here is that it's hard to determine when a function signature may be implied that an `{String id}` belongs to that particular service-oriented entity. For instance, there are a fetch in memoservices that returns all memos, but one with the collection Id and one that return a single one with the unique id for that particular memo. In this case, it clearly doesn't specify that the ID belongs collection, so the argument being more verbose is intended, otherwise, I would say that this signature should be `getCollectionMemoryStabilityByCollectionId(String id)` which, in my opinion is a little bit worse, but debatable

lib/domain/services/collection_services.dart
matuella: Actually, counting only pristine versus all memos are equally bad, because they should never be used in the application unless the collection has all memos executed, so, even though this still returns a value, it may not  mean much, given the fact that only fully completed collections should be evaluated

lib/domain/services/execution_services.dart
matuella: It's a good suggestion, although we still need to keep reference to the count of the `executedMemoPerStability`, because even though the chunk require, let's say, 50 memos, and the total collection has less than the chunk, it would stack overflow in the while condition. I will make the according changes

lib/domain/services/execution_services.dart
matuella: This is a problem that I will have addressed in a future PR, but it's a valid proposition

lib/domain/services/execution_services.dart
ggirotto: So considering that this function is always called right before an execution, and never during it, what do you think about my suggestion of updating the function docs to emphasis that this chunk of memos will be used in a execution that is to come instead using `when executing`?

lib/domain/services/collection_services.dart
ggirotto: I was remembering now from our discussion about only show the memory stability when all collection cards were finished. With this definition in mind, having memos in this function without stability isn't an inconsistency? In theory, all of them should have a stability at this function

matuella:

PR #51, opened 2021-04-29
All versioning related features

matuella:


lib/data/repositories/collection_repository.dart
```diff
@@ -62,17 +73,39 @@ class CollectionRepositoryImpl implements CollectionRepository {
...
+
+    final castCollections = List<Map<String, dynamic>>.from(rawCollections);
+    return castCollections.map(_collectionsMemosSerializer.from).toList();
+  }
 
-    final listenFutures =
-        collectionIds.map((collectionId) => _db.listenTo(id: collectionId, store: _collectionStore)).toList();
-    final rawCollectionListeners = await Future.wait(listenFutures);
+  @override
+  Future<void> putCollectionsWithCollectionMemos(List<CollectionMemos> collections) async {
```
ggirotto, line 85: This parameter shouldn't be `memos` or `collectionMemos` instead `collections`?

lib/data/repositories/collection_repository.dart
```diff
@@ -62,17 +73,39 @@ class CollectionRepositoryImpl implements CollectionRepository {
...
+    return _db.putAll(
+      ids: collections.map((collection) => collection.id).toList(),
+      objects: collections
+          .map(
+            (collection) => {
```
ggirotto, line 90: Isn't this the work of the `CollectionSerializer`?

lib/domain/models/memo_collection_metadata.dart
```diff
@@ -0,0 +1,21 @@
+/// Fundamental fields that defines a `Memo` when relating to a `Collection`
+class MemoCollectionMetadata {
+  MemoCollectionMetadata({required this.uniqueId, required this.rawQuestion, required this.rawAnswer});
+
+  /// Identifies a global unique id (unique both in the parent's `Collection` and through all other `Memo`)
+  final String uniqueId;
```
ggirotto, line 6: I confess that I'm a little bit confuse about this `unique` usage. By default, identifiers should be treated as unique, why do we need to specify that it's unique?

lib/data/repositories/memo_repository.dart
```diff
@@ -7,8 +7,20 @@ abstract class MemoRepository {
   /// Retrieves all available [Memo] that belongs to a `Collection` with id [collectionId]
   Future<List<Memo>> getAllMemos({required String collectionId});
 
-  /// Batch update a list of [memos]
-  Future<void> updateMemos(List<Memo> memos);
+  /// Retrieves all available [Memo] that belongs to one any of the `Collection`s with [collectionIds]
```
ggirotto, line 10: one any

lib/domain/models/collection.dart
```diff
@@ -64,3 +69,22 @@ class Collection extends MemoExecutionsMetadata with EquatableMixin {
         ...super.props,
       ];
 }
+
+/// Represents all `Collection` metadata
+abstract class CollectionMetadata {
```
ggirotto, line 74: Why this abstraction?

ggirotto: I couldn't find other entity that benefits of it

lib/domain/models/memo_collection_metadata.dart
```diff
@@ -0,0 +1,21 @@
+/// Fundamental fields that defines a `Memo` when relating to a `Collection`
+class MemoCollectionMetadata {
```
ggirotto, line 2: Shouldn't this be abstract?

ggirotto: Also, I would like to raise a discussion about this naming. These metadata are related to a memo, and they are used by `Memo` and `MemoExecution` entities correct? If so, why don't call it just `MemoMetadata`?

lib/domain/transients/collection_memos.dart
```diff
@@ -0,0 +1,41 @@
...
+  @override
+  final String name;
+  @override
+  final String description;
+  @override
+  final String category;
+  @override
+  final List<String> tags;
+  @override
+  int get uniqueMemosAmount => memosMetadata.length;
```
ggirotto, line 32: Shouldn't have a line break between these properties? Does lint accepts this style?

ggirotto (changes requested):

ARCHITECTURE.md
```diff
@@ -273,6 +277,10 @@ structures (called Use Cases in the clean architecture) if proven necessary.
 They are the only structures in [`domain/`](#domain) that communicates with the [`data/`](#data) layer, more
 specifically, through the [`repositories/`](#repositories).
 
+#### `transients/`
+
+A domain model - a set of structures that represent a business object.
```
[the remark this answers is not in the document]
matuella: Not sure why my description got overwritten, it wasn't a copy paste from the `models`, I'm not quite sure about this description, but I will obviously update mine

lib/application/view-models/app_vm.dart
```diff
@@ -19,27 +24,39 @@ import 'package:sembast/sembast.dart';
...
 }
 
 class AppVMImpl extends AppVM {
-  AppVMImpl() : super(const AsyncValue.loading()) {
-    _loadAppVM();
-  }
+  AppVMImpl() : super(const AsyncValue.loading());
+
+  /// Manages if the dependencies have already been requested to load
+  var _hasRequestedLoading = false;
```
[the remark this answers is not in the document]
matuella: but that is exactly what it means, we just use it to make sure that there is a single request for loading,  the act of defining being loaded is done by the `AsyncValue` object (`AppVMImpl`).

lib/data/gateways/sembast_database.dart
```diff
@@ -118,9 +137,9 @@ class SembastDatabaseImpl implements SembastDatabase {
 
   /// Transforms a single `sembast` snapshot record into an object
   final snapshotTransformer =
-      StreamTransformer<RecordSnapshot<String, Map<String, Object?>>, Map<String, Object?>>.fromHandlers(
+      StreamTransformer<RecordSnapshot<String, Map<String, Object?>>?, Map<String, Object?>?>.fromHandlers(
     handleData: (snapshot, sink) {
-      sink.add(snapshot.value);
+      sink.add(snapshot?.value);
```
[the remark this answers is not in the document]
matuella: That's because it retrieves an object from the database, and it may be simply `null`. The listeners are the ones that have to deal with it being null or not, and not the stream transformer, because there might be some scenarios where an object simply is deleted, so an stream that did return a non-null object may start returning a null object, and vice-versa

lib/data/repositories/collection_repository.dart
matuella: That's actually confusing, but couldn't finda better naming, so no, it's a CollectionMemos instead, a list of a collection glued with a list of their respective memos metadata (this is a transient class)

lib/domain/models/memo_collection_metadata.dart
matuella: Because we must reinforce this through the naming. One might think that a memo id could be unique only between its siblings in a collection, but no, it must be globally unique between all other memos

lib/data/repositories/collection_repository.dart
matuella: Nope, it only updates part of the object. This is a very useful feature because one may not need to fetch an object before updating, so we can do partial updates and in a type-safe fashion, where we expose the fields by the keys (in this case the CollectionKeys)

lib/domain/models/collection.dart
matuella: It's used by both `CollectionMemos` and `Collection`, required because we must deal with very similar, but inherently different collection objects, one that is store locally and one that is stored in sembast

lib/domain/models/memo_collection_metadata.dart
matuella: No, they are also instantiated as a list of local memo associated with a local collection file (.json), in `CollectionMemos` transient

lib/domain/models/memo_collection_metadata.dart
ggirotto: IMO you're over engineering this definition. It's easier to assume that all identifiers are unique by default, instead reinforcing it in the property name and in its description. As an example, the collections identifiers are also unique, but their id property stands only by `id`. If we someday decide to wrap the collections in another structure that holds a set of collections, we would not refactor this `id` property to reinforce that it's unique. We could add a note on README or somewhere else about this definition that all ids are unique, but in code will create a unnecessary complexity IMO

lib/domain/models/memo_collection_metadata.dart
ggirotto: But this class has no relation with a Collection on its scope. It has an id, a question and an answer. I don't see why having Collection on its name. Even on CollectionMemos class, its responsibility is to hold the memos metadata. Although it can be used in a context that has relation to a Collection, the class scope itself doesn't have any reference to a collection, as far as I can see

lib/domain/models/memo_collection_metadata.dart
ggirotto: IMO you're over engineering this definition. It's easier to assume that all identifiers are unique by default, instead reinforcing it in the property name and in its description. As an example, the collections identifiers are also unique, but their id property stands only by id. If we someday decide to wrap the collections in another structure that holds a set of collections, we would not refactor this id property to reinforce that it's unique. We could add a note on README or somewhere else about this definition that all ids are unique, but in code will create a unnecessary complexity IMO

ggirotto (approved):
@matuella My last two statements stay, but I'm approving this PR since they are more a personal opinion than anything elese. I added all my thoughts about those naming decisions but I'll let you decide, since you are in the project development frontline. You can mention me if you want to discuss more about them 💪

matuella:

PR #52, opened 2021-04-29
All home changes given the new execution functionality

matuella:


lib/application/constants/strings.dart
```diff
@@ -16,18 +16,51 @@ const collectionsExploreTab = 'Explorar';
...
 const collectionsSectionHeaderSeeAll = 'Ver todos';
-const collectionsMemoryStability = 'Estabilidade da Memória';
-String collectionsCompletionProgress({required int current, required int target}) =>
-    '$current / $target memos completados';
 
-const progressTotalStudyTime = 'Horas totais de estudos';
+//
+// Progress
+//
+const progressTotalStudyTime = 'Totais de estudos';
```
ggirotto, line 23: It shouldn't be `estudo` instead `estudos`? I'm not sure

lib/application/constants/strings.dart
```diff
@@ -16,18 +16,51 @@ const collectionsExploreTab = 'Explorar';
...
+
+const executionYourPerformance = 'Seu desempenho';
+const executionBackToCollections = 'Voltar para as coleções';
+
+const executionWellDone = '## Muito Bem';
+const executionImprovedKnowledgeDescription = 'Você acaba de aprimorar seu conhecimento em:';
 
-String progressDifficultyEmoji(MemoDifficulty difficulty) {
+String executionLinearIndicatorCompletionLabel(String completionDescription) =>
+    'Indicador linear demonstrando que o nível de conclusão desta sessão aprendizado está em $completionDescription';
```
ggirotto, line 42: desta sessão **de** aprendizado está em

lib/application/coordinator/routes.dart
```diff
@@ -52,6 +52,33 @@ class ProgressPath extends HomePath {
   String get formattedPath => '/$name';
 }
 
+//
+// Collection Details
+//
+class CollectionDetailsPath extends AppPath {
+  CollectionDetailsPath(this.collectionId);
+  static const name = 'collection_details';
```
ggirotto, line 60: What do you think about creating a routes file which holds all routes strings?

ggirotto (changes requested):

lib/application/constants/strings.dart
matuella: I think both are correct

lib/application/coordinator/routes.dart
matuella: I don't think it's useful, as they are only used here, and they are already in their own routes

ggirotto (approved):

matuella:

PR #53, opened 2021-04-29
Updates to connect the execution view model with the new services approaches

matuella:


lib/application/view-models/execution/collection_execution_vm.dart
```diff
@@ -3,70 +3,165 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
...
 
 abstract class CollectionExecutionVM extends StateNotifier<CollectionExecutionState> {
   CollectionExecutionVM(CollectionExecutionState state) : super(state);
 
   /// Marks the current [Memo] with the [difficulty]
   ///
-  /// This mark will always trigger a new [LoadedCollectionExecutionState] update with the next [Memo] and, if there are
-  /// no more [Memo]s, the new state should be a [FinishedCollectionExecutionState].
-  Future<void> markCurrentMemoDifficulty(MemoDifficulty difficulty);
+  /// This action of marking will always trigger a new [LoadedCollectionExecutionState] update with the selected
```
ggirotto, line 25: I would go with `It always trigger ....`

lib/application/view-models/execution/collection_execution_vm.dart
```diff
@@ -3,70 +3,165 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
...
+  /// [difficulty]
+  ///
+  /// Throws an [InconsistentStateError] if the state is other than [LoadedCollectionExecutionState]
+  void markCurrentMemoDifficulty(MemoDifficulty difficulty);
+
+  /// Go forward with the next suitable contents
+  ///
+  /// There are two scenarios where we should allow an execution to go forward:
+  ///   1. there is a question being displayed and it should provide the question to the current memo; or when
```
ggirotto, line 34: What does it mean that the question should be provided? I mean, the sentence says:

> there is a question being displayed

and then it says:

> it should provide the question

How the question is being displayed but it needs to be provided? If this is related to the question content, I would change it to something like:
`There is a question waiting to be displayed and it should provide this question content to the memo`

lib/application/view-models/execution/collection_execution_vm.dart
```diff
@@ -3,70 +3,165 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
...
+
     final loadedState = state as LoadedCollectionExecutionState;
 
+    // If there is a `markedAnswer`, we assume that this is a request to forward with the last marked difficulty
+    if (loadedState.markedAnswer != null) {
+      await _confirmMarkedMemo(loadedState.markedAnswer!);
+    } else {
+      // Otherwise, we expect it to be a request to display the question
+      final currentMemo = _memos[_executions.length];
+      state = loadedState.copyWith(isDisplayingQuestion: false, currentContents: currentMemo.rawAnswer);
```
ggirotto, line 86: The doc says that this is a request to display the question, but the state copy is displaying the answer

lib/application/view-models/execution/collection_execution_vm.dart
```diff
@@ -3,70 +3,165 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
...
-      state = FinishedCollectionExecutionState();
+      await executionServices.addExecutions(_executions, collectionId: collectionId);
+
+      final mappedExecutions = <MemoDifficulty, int>{
+        for (final difficulty in MemoDifficulty.values) difficulty: 0,
+      };
+
+      _executions.forEach((execution) {
+        final difficultyValue = mappedExecutions[execution.markedDifficulty];
+        mappedExecutions[execution.markedDifficulty] = difficultyValue! + 1;
```
ggirotto, line 117: You probably tried it, but cant we just make `mappedExecutions[execution.markedDifficulty] += 1`?

ggirotto (changes requested):

lib/application/view-models/execution/collection_execution_vm.dart
matuella: we sadly can't, but I thinlk there is a possibility of doing it by extending the operators, although I haven't tried

lib/application/view-models/execution/collection_execution_vm.dart
matuella: I've updated the docs, it should always skip to the next suitable set of contents, meaning an answer or a question, no matter which order, the vm knows how to handle

lib/application/view-models/execution/collection_execution_vm.dart
```diff
@@ -22,19 +22,19 @@ abstract class CollectionExecutionVM extends StateNotifier<CollectionExecutionSt
...
   /// There are two scenarios where we should allow an execution to go forward:
-  ///   1. there is a question being displayed and it should provide the question to the current memo; or when
-  ///   2. there is an answer *with a marked answer* and it should proceed to next question (or change its state to a
-  /// [FinishedCollectionExecutionState] if there are no more memos left). In this case, if `nextContents` is requested
-  /// displaying an answer with no marked answer, it won't do anything.
+  ///   1. there is a question being displayed and it should provide the answer for this question; or when
+  ///   2. there is an answer contents being displayed, **with the according marked difficulty**, and it should proceed
+  /// to next question (or change its state to a [FinishedCollectionExecutionState] if there are no more memos left). In
+  /// this same case, if `nextContents` is requested while displaying an answer with no marked answer, it won't do
+  /// anything.
```
ggirotto, line 37: Way better 👍

ggirotto (approved):

matuella:

PR #58, opened 2021-04-29
Collection Execution changes + Serializer Tests Updates

matuella:


lib/domain/models/memo.dart
```diff
@@ -10,6 +10,7 @@ import 'package:meta/meta.dart';
 /// infinite amount of times, its purpose is to store the most recent version of its question/answer with useful
 /// execution's metadata, like [lastExecution] and by extending [MemoExecutionsMetadata].
 @immutable
+// ignore: avoid_implementing_value_types
```
ggirotto, line 13: Is this rule necessary?

ggirotto (approved):

lib/domain/models/memo.dart
matuella: I think they are really useful in scenarios one can't run away from mutable-forced classes, like this one

matuella:
All updates made and merged in the last PR (#59) in this nested chain of collection execution tests PRs

PR #62, opened 2021-05-01
Release Configuration + Fastlane + CI

ggirotto:


ios/fastlane/Changelog.txt
```diff
@@ -0,0 +1 @@
+Bugs fixing and performance improvements
```
ggirotto, line 1: This CHANGELOG file can be updated to add custom information to external testers

.github/workflows/pull-request.yml
```diff
@@ -1,4 +1,4 @@
-name: Flutter CI
+name: Pull Request Workflow
```
matuella, line 1: It's on both PR and push. While you are at it, why we don't default only to PR? We don't honestly care if an individual commit breaks anything (or fail tests), we only care for branches and its respective pull requests being made to the `main` anyways.

.github/workflows/release.yml
```diff
@@ -0,0 +1,58 @@
...
+  build:
+    runs-on: macos-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v1
+
+      - name: Setup Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: "12.x"
```
matuella, line 18: Why exactly is java needed for this actions?

.github/workflows/release.yml
```diff
@@ -0,0 +1,58 @@
...
+      # are used to sign the Android app. Then run `internal` lane to build and deploy Android app
+      - name: Release Android App
+        run: |
+          echo "${{ secrets.KEYSTORE_BASE64_ENCODED }}" > keystore.jks.asc
+          gpg -d --passphrase "${{ secrets.KEYSTORE_PASHPHRASE }}" --batch keystore.jks.asc > android/app/keystore.jks
+          echo "${{ secrets.KEY_PROPERTIES_BASE64_ENCODED }}" > key.properties.asc
+          gpg -d --passphrase "${{ secrets.KEY_PROPERTIES_PASSPHRASE }}" --batch key.properties.asc > android/key.properties
+          cd android && fastlane internal --verbose
+        env:
+          GOOGLE_SERVICE_ACCOUNT: ${{ secrets.GOOGLE_SERVICE_ACCOUNT }}
```
matuella, line 58: I think you could explain a little more in-depth of actually goes around here in the ARCHITECTURE file, more specifically about this `ssh-key-action` dependency

ios/Runner/Info.plist
```diff
@@ -10,18 +10,22 @@
...
 	<string>$(FLUTTER_BUILD_NAME)</string>
 	<key>CFBundleSignature</key>
 	<string>????</string>
 	<key>CFBundleVersion</key>
-	<string>$(FLUTTER_BUILD_NUMBER)</string>
+	<string>2</string>
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
+    <key>ITSAppUsesNonExemptEncryption</key>
+    <false/>
```
matuella, line 28: What is the reason for this?

matuella (commented):
So, I understand that this automatically uploads a new version to both stores (in a test environment and not straight to production), but I'm not sure if this action:
- will bump AND COMMIT to this repository when a new release is successful?
- will it reset AND COMMIT to this repository if a new version (major, minor or patch) is updated?
- what happens if one of the builds fail (because we know it's impossible to guarantee atomicity in these two operations)? What is our fallback here? Could it retry?

I think you could go address these particularities in the ARCHITECTURE document as well.

Also, I'm not 100% sure, but I can't remember if we can't use the flutter's build number because of a limitation in fastlane or in the flutter ecosystem itself. Here I am assuming that we DO want the same build version for both iOS and Android, and that we can reset the build number for the Android's ecosystem.

.github/workflows/pull-request.yml
ggirotto: Agreed

RaphaelOliveiraMoura (approved):

.github/workflows/release.yml
ggirotto: Flutter requires Java to build the Android app, even if it is built using Kotlin

ios/Runner/Info.plist
ggirotto: Every time we submit an iOS app to review, Apple ask if the app is using any kind of encryption beyond those that are exempt from the export compliance. Since we don't use, and not even plan to use, I added this property to Info.plist indicating that we don't use encryption to avoid Apple asking for encryption every time we submit a new build

See for more info: https://developer.apple.com/documentation/bundleresources/information_property_list/itsappusesnonexemptencryption

ggirotto:
@matuella 

> will bump AND COMMIT to this repository when a new release is successful?

It wasn't. Fixed on my last commits.

> will it reset AND COMMIT to this repository if a new version (major, minor or patch) is updated?

The build number is controlled by Fastlane lanes, but the version bump is our responsibility. It will not update the version automatically, just bump the build number

> what happens if one of the builds fail (because we know it's impossible to guarantee atomicity in these two operations)? What is our fallback here? Could it retry?

If any phase of the scripts fails, it doesn't execute the next phases. It means that, it may successfully update the iOS app but fail in the Android lane, as an example. We don't have a fallback in the script itself. If it failed because an error in the workflow itself, we need to fix it and then submit a new PR pointing to `main` with the fix, and it will naturally trigger the release workflow and repeat the process. If it failed due to an external source, like a connection issue with the stores, we just need to re-run the workflow through [Actions tab](https://github.com/olmps/memo/actions), since the workflow script itself doesn't have any kind of retry workaround or stuff like that.


> I think you could go address these particularities in the ARCHITECTURE document as well.

I added a few docs on ARCHITECTURE.md pointing to the release lanes and Fastlane files. I didn't write much about how the release workflow and the Fastlane files work because all of the mare well documented and are kind of self-explanatory, so I kept the explanation in the ARCHITECTURE.md simpler.

> Also, I'm not 100% sure, but I can't remember if we can't use the flutter's build number because of a limitation in fastlane or in the flutter ecosystem itself. Here I am assuming that we DO want the same build version for both iOS and Android, and that we can reset the build number for the Android's ecosystem.

Yes, we are not using Flutter's build number defined in `pubspec.yaml`, we are using those specified in the platforms settings (build.gradle for Android and PBX for iOS). We use `pubspec.yaml` to retrieve the version code only (major.minor.patch), but the build number are specified and managed directly in the platform source files.

`iOS` Fastlane has an additional feature that it sets as the build number the latest build number available on testflight + 1. Android just bumps the build number that exists in the local `build.gradle`, even if this build number is not in Google Play store yet. This because iOS has an action that supports fetching this information from App Store Connect, while Android doesn't.


> we can reset the build number for the Android's ecosystem

We cannot reset the build number for Android apps, only in iOS. It means that the Android build number will increment indefinitely, while iOS will reset on every version change.

.github/workflows/release.yml
ggirotto: Agreed

matuella (changes requested):
Thanks for the changes.

There is still one thing that I think it's relevant here, and it's about adding an overview what we are doing with how the fastlane + github actions tools.

Adding a description of how the build increment, the auto-commit, the errors fallbacks (or how would we do if it fails), and how all of this work in conjunction is still very useful and provides value to the reader, otherwise one would have to dig in all files, workflows and such, just to get an overview what we are doing.

Your argument about it being self-explanatory is contradictory to why we explain the architecture and all the blablabla, although your case is correctly assumed way more superficial than the architecture. But can you see that the same inference can be done with the architecture just by going through the files? The point of explaining it in the ARCHITECTURE gives much more context to one that is trying to understand what is going on, so even though it might be one or two phrases, it may provide a lot of "basis" to someone that is new to these tools.

ggirotto:
@matuella Agreed. Please read the new docs on `ARCHITECTURE.md` and let me know what you thought about it. If you missed some information, please let me know because I'm biased since the script was written by me, so everything looks obvious in my point of view

ARCHITECTURE.md
```diff
@@ -445,4 +445,31 @@ And that's it, the currently supported environments are: `DEV` and `PROD`.
...
+[Fastlane](https://fastlane.tools/) automation tool. Fastlane gives us a set of tools to automate all the heavy work
+of configuring a project environment to be deployed, build and deploy it to production. Every the release script from
+Github Actions finish the environment setup, we use Fastlane to run the targets tests, analyzer, bump the build version
+from the targets, build the projects and finally deploy them. We currently have Fastlane running for iOS and Android 
+targets, and the lanes that perform such tasks are described in details in both platforms Fastfile's:
+
+- [iOS Fastfile](ios/fastlane/Fastfile)
+- [Android Fastfile](android/fastlane/Fastfile)
+
+The main difference between the mobile platforms, is that Android build number is indefinitely incremented, since it's
```
matuella, line 466:
```suggestion
The main difference between the mobile platforms is that Android build number is indefinitely incremented, since it's
```

matuella (approved):

PR #65, opened 2021-05-04
Adds `scaffoldMessenger`

matuella:


pubspec.yaml
```diff
@@ -21,6 +21,7 @@ dependencies:
   equatable: ^2.0.0
   path: ^1.8.0
   path_provider: ^2.0.0
+  url_launcher: ^6.0.3
```
ggirotto, line 24: Why this package was added on this PR? It's not used in any of the introduced files. Also, this package advises some extra setup in iOS to work properly as far as I understood: https://pub.dev/packages/url_launcher#installation

ggirotto (changes requested):

pubspec.yaml
ggirotto: I found its usage in this future PR: https://github.com/olmps/memo/pull/69/files but it doesn't have the iOS configuration. that the package describes in the tutorial. Are these steps optional?

matuella: yeah, it should've been added in #69

Not sure about this config, on emulator it worked fine. Sadly, can't currently test on a physical device atm, but will keep this in mind before merging all these multi-PR changes

ggirotto (approved):

matuella:
All updates made and merged in the last PR (#76) in this nested chain of collection execution PRs

PR #74, opened 2021-05-04
Adds `CollectionDetailsPage` and `CollectionDetailsVM`

matuella:


lib/application/view-models/details/collection_details_vm.dart
```diff
@@ -0,0 +1,104 @@
...
+  final CollectionServices collectionServices;
+  final ResourceServices resourceServices;
+
+  List<Resource>? _associatedResources;
+  late final StreamSubscription<CollectionStatus> _listener;
+
+  Future<void> _loadCollection() async {
+    final stream = await collectionServices.listenToCollectionStatus(collectionId: collectionId);
+
+    _listener = stream.listen((collectionStatus) async {
```
ggirotto, line 53: I couldn't understand why the details vm listen to the collection status. What's the scenario where the collection status may change while this page is loaded in memory?

ggirotto (changes requested):

lib/application/view-models/details/collection_details_vm.dart
matuella: the user might return to this page, for any possible reason, after executing the same collection

ggirotto: Ahh, got it. The execution is presented in a modal

ggirotto (approved):

matuella:
All updates made and merged in the last PR (#76) in this nested chain of collection execution PRs

Issue #39, opened 2021-04-10
Codecov is not working properly


matuella:
You can see that in the PRs #35, #36, #37 and #38 , the coverage suggestion is for some reason the same as the #33

RaphaelOliveiraMoura:
@matuella Does this problem keep happening?

matuella:
It appears that this only occurs when multiple PRs are opened simultaneously. I will try to replicate this issue once again, if not, I will close it. Thanks for asking!

abelborges:
FYI https://arstechnica.com/gadgets/2021/04/backdoored-developer-tool-that-stole-credentials-escaped-notice-for-3-months/

matuella:
@abelborges thanks for the info, I have been watching these codecov issues and I think this isn't a reliable tool anymore for us to use with github actions, which could lead to use eventually exposing some private keys (through a supply chain attack).

https://www.theregister.com/2021/05/05/twilio_codecov_attack/ is another good ref.

My take on this is that we should migrate to [coveralls](https://coveralls.io/) or [code climate](https://codeclimate.com/).

matuella:
Given the new updates made to the github actions, I am closing in favor of #84

PR #78, opened 2021-05-04
Adds Fastlane + Github Actions to auto-deploy new releases (#62 clone)

ggirotto:
https://github.com/olmps/memo/pull/62 was merged in the wrong branch `collection-execution`. This PR addresses the same changes already discussed and reviewed at https://github.com/olmps/memo/pull/62 but now pointing correctly to `main`, since `collection-execution` was already merged


matuella:
Why not simply reverse?

ggirotto:
The revert tool from Github creates a new PR reverting the changes from the merged PR. Since the `collection-execution` was already merged on `main`, its easier to just create a new PR pointing to `main` since eventually `collection-execution` will be deleted anyways, we don't need to revert the merge

matuella (approved):

ggirotto:
@matuella The pull-request workflow wasn't working anymore. I refactored it by adding Fastlane to run the build lanes. Please review the last commits again

.github/workflows/pull-request.yml
```diff
@@ -6,41 +6,67 @@ on:
     branches:
```
ggirotto, line 6: @matuella You may notice that most of the workflow is copy-pasted from `release` workflow. Github Actions has "composite actions" while let us reuse the workflow scripts, but it currently has a lot of limitations that makes unsustainable to use it as it is today

See more: https://github.com/actions/runner/issues/646

matuella: Why don't we create a blocked issue that can eventually be solved when this runner issue has been solved?

matuella: Also, why do we exactly need to setup the release credentials on a PR check? Why is fastlane called here?

.github/workflows/release.yml
```diff
@@ -0,0 +1,87 @@
...
+          git config --global user.name 'Memo Bot'
+          git config --global user.email 'memo@olmps.co'
+          git commit -am "[Fastlane] Bump version"
+          git push
+
+      # Run tests and upload code coverage do Codecov
+      - name: Run tests and upload coverage to Codecov
+        run: |
+          flutter test --coverage
+          curl -s https://codecov.io/bash | bash
```
matuella, line 87: don't mix codecov with our private stuff, if you have ran this, we could have possibly leaked the keys already. Read more #39

android/app/build.gradle
```diff
@@ -33,19 +34,25 @@ android {
     }
 
     defaultConfig {
-        // TODO: Specify your own unique Application ID (https://developer.android.com/studio/build/application-id.html).
-        applicationId "com.example.memo"
+        applicationId "com.olmps.memoClient"
```
matuella, line 37: that's a... non standard application bundle id naming you know

ios/Runner/Info.plist
```diff
@@ -10,18 +10,22 @@
 	<string>$(PRODUCT_BUNDLE_IDENTIFIER)</string>
 	<key>CFBundleInfoDictionaryVersion</key>
 	<string>6.0</string>
+  <key>NSPhotoLibraryUsageDescription</key>
+  <string>We need to access your photo library so you can choose your own pictures when creating your Memos</string>
```
matuella, line 14: why is this string repeated here and in en.lproj?

matuella (changes requested):

ios/Runner/Info.plist
ggirotto: This one is the fallback if the user is running in a locale that doesn't have a dedicated localization file

.github/workflows/release.yml
ggirotto: I removed the Codecov phase until we decide how to proceed with this security issues they are facing. Further discussions will be made through https://github.com/olmps/memo/issues/39

.github/workflows/pull-request.yml
ggirotto: > Why don't we create a blocked issue that can eventually be solved when this runner issue has been solved?

Done: https://github.com/olmps/memo/issues/80

> Also, why do we exactly need to setup the release credentials on a PR check? Why is fastlane called here?

You're right, we shouldn't have secret api keys when building

matuella (changes requested):
As we have discussed in Discord, fastlane is not a necessity in the PR workflow

ggirotto:
@matuella Removed the iOS and Android Fastfile's build lanes (since they are not used anymore) and updated pull-request lane removing the usage of secret api keys and tokens (disabled iOS and Android code signing)

matuella (approved):
Overall, really nice addition!

Just one other point, is there really no alternative to that little "hacking" in Android?

.github/workflows/release.yml
```diff
@@ -15,6 +15,8 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
+        with:
```
ggirotto, line 18: We need to specify the branch during checkout because this script is triggered from a tag publish, and we need to checkout to the correct branch because the last step is to commit the bumped files

ggirotto:
@matuella I was testing both workflows before merging and found two inconsistencies. I made the necessary updates but I need that you review this last changes

matuella (approved):

PR #79, opened 2021-05-05
UI fixes and enhancements

matuella:
Fixes and enhancements while testing the first release candidate.

## Enhancements:
- Adds ptBR FAQ;
- Adds resources warning string in `CollectionDetailsPage`;
- Adds recall explanation link in completed executions
- Adds new help + faq section in settings;
- Updates `CollectionsPage` with empty states;
- Adds a `AlertDialog` before dismissing the current execution.

## Fixes:
- Adds missing configuration for `url_launcher` to work properly;
- Adds missing themed containers in `CollectionDetailsPage` and `CollectionsPage`;
- Adds missing navigation icon for `NamedCustomSettingsItem`;
- Adds missing opacity to unselected terminal actions.


FAQ_ptBR.md
```diff
@@ -0,0 +1,31 @@
...
+Uma coleção se trata de um grupo de [Memos](#o-que-é-um-memo).
+
+Fazendo uma analogia com [flashcards](https://en.wikipedia.org/wiki/Flashcard), ela é o que consideramos um Deck.
+
+## O que é um Memo?
+
+Um memo é um "desafio" individual de uma [Coleção](#o-que-é-uma-coleção).
+
+Fazendo uma analogia com [flashcards](https://en.wikipedia.org/wiki/Flashcard), é o que consideramos como uma Carta,
+esta na qual pertence à um Deck.
```
ggirotto, line 19: `esta na qual`? It doesn't seem correct to be. Wouldn't it be just `a qual...`?

lib/application/widgets/theme/link.dart
```diff
@@ -96,31 +99,71 @@ class ExternalLinkButton extends StatelessWidget {
...
-      }
-    } on PlatformException catch (exception) {
-      final urlException = URLException.failedToOpen(debugInfo: exception.toString());
-      onFailLaunchingUrl?.call(urlException);
+Future<void> _handleUrlLaunch(String url, Function(UrlException exception)? onFailLaunchingUrl) async {
+  try {
+    if (await url_launcher.canLaunch(url)) {
+      await url_launcher.launch(url);
+    } else {
+      onFailLaunchingUrl?.call(UrlException.failedToOpen());
```
ggirotto, line 163: Ins't valid a debug info adding the URL that couldn't be opened to the exception? It will be easier to debug if we know the url

lib/application/widgets/theme/themed_bottom_container.dart
```diff
@@ -0,0 +1,50 @@
+import 'package:flutter/material.dart';
```
ggirotto, line 1: This file name could be just `themed_container` instead `themed_bottom_container`, since it has both: `top` and `bottom`

FAQ_ptBR.md
```diff
@@ -0,0 +1,31 @@
...
+
+## O que é o nível de fixação?
+
+O nível de fixação (traduzido de *memory recall*) pode ser definido como **a soma das estimativas que damos à
+probabilidade de um usuário ter memorizado todos os Memos de uma Coleção**. Este nível é calculado com base na
+[curva de esquecimento do SuperMemo](https://supermemo.guru/wiki/Forgetting_curve).
```
ggirotto, line 25: What about instead
> **a soma das estimativas que damos à probabilidade de um usuário ter memorizado todos os Memos de uma Coleção**

We say like:
> **a soma das forças de memória do usuário para cada Memo, isto é, o quão forte é a memória do usuário sobre os Memos de uma Coleção**

SuperMemo uses the term [*Memory Strength*](https://supermemo.guru/wiki/Two_components_of_memory), which I personally think that makes a lot of sense

ggirotto (changes requested):

FAQ_ptBR.md
matuella: But that's not what we estimate, we only use the recall given a forgetting curve. What we calculate is the `R - probability of recall (retrievability of memory)`, and not the strength of a memory. They may be similar, but I prefer to maintain the nomenclature defined by SuperMemo, which is our core reference for this implemenation.

matuella: I will update given your suggestion nonetheless, given that the overall syntactic structure is better

lib/application/widgets/theme/link.dart
matuella: i added it but forgot to pass it

FAQ_ptBR.md
matuella: I don't think so. In this case, I specifically refer to "esta na qual" as the "Carta" reference.

ggirotto (approved):

PR #82, opened 2021-05-06
Assets testing

matuella:
- Added an even more type-safe approach to `images`, `fonts` and `licenses`;
- Added a generic test for all assets in folder, meaning that we won't have runtime issues with all assets used;
- Added tests for non-standard assets, meaning, the collections, resources and memos, so we can check out any inconsistency in all of our versioned models, meaning resources, collections and memos.


ARCHITECTURE.md
```diff
@@ -433,6 +434,22 @@ experience for offline users;
 The only drawback of doing this manually is importing the fonts in the `assets/` folder (size shouldn't be relevant to
 impact even low-end devices), and specifying them in the `pubspec.yaml`.
 
+## Importing Fonts
```
ggirotto, line 437: This docs could also mention the image assets, since the approach is the same right?

ggirotto (approved):
Nice improvements 💪

ARCHITECTURE.md
matuella: I don't think image is useful, because it's pretty much straightforward, contrary to fonts, which has this specific naming that I had to highlight, because the only way to know about this is to actually go into `fonts_test.dart` and check it out a utility function

PR #83, opened 2021-05-06
Adds `TransactionHandler` and database transaction

matuella:


ggirotto (approved):
Nice feature. We need to find a way to run multiple transactions in the current architecture. I suspect that exists a simple solution, but I couldn't find it yet

matuella:
@ggirotto 
> Nice feature. We need to find a way to run multiple transactions in the current architecture. I suspect that exists a simple solution, but I couldn't find it yet

I only see this case where multiple services are doing simultaneous transaction operations and these should be a really rare case because, usually, our view models only access a single service instance to do all of its work, as they are considered an aggregation of the clean architecture's use cases. 

Nonetheless, these multiple-services-transactions are completely _workaroundable_ by simply waiting one transaction finish before calling the other, in a synchronous way.

What usually happens in a transaction is a lock (or mutex) that doesn't allow any other changes while the "transaction" changes stuff so, virtually they never run in parallel, as there are conflicts and deadlocks to be handled, and we will be diving into a SQL concept, which is not what we are exactly aiming here, because sembast deals with all of its logic in-memory, which makes these discussions of ACID stuff even more abstract.

What one could do is actually queue such transactions, but I think the work doesn't pay off, giving that we are talking about a client application here that doesn't deal with an exaggerated amount of data.

So, IMO, we don't need to find a way to run multiple transactions simultaneously.

ggirotto:
@matuella Completely agree. It doesn't make sense to me to have such hard work to execute in a client app. I'm just wondering if we someday find a specific use-case that need such heavy work, how could we fit this parallel transactions in a client app. But I think that in this scenario we would have to instantiate multiple DB instancies and services.

Issue #94, opened 2021-05-12
Improve speed in workflow build-related operations


matuella:
Not sure if this is a limitation of the images that we are using or what, but both release and PR workflows are slow. ​

What could we do to improve these build times (possibly [caching](https://github.com/actions/cache)?

matuella:
@ggirotto what do you think about this?

ggirotto:
I have a strong hypothesys about speeding up the `release` workflow. Today we build the iOS app twice: first by using `flutter build` and second by using `build_ios_app` fastlane action. I'm almost sure that we could just remove the `build_ios_app` fastlane action and keep just `flutter build`. I've notes on my side to test it, but I haven't tested yet.

In the `pull_request` workflow, we could definitely cache flutter SDK. I've tried for a few hours, but without success I gave up because this github actions task were taking to long, but now I think we could go back and improve this caching stuff
Cache reference: https://github.com/subosito/flutter-action/issues/16

matuella:
I think #98 resulted in a decent performance improvement (I guess it was around ~33% of less time spent passing the workflow).

@ggirotto Are there other improvements that comes to your mind in both `pull_request` and `release` workflows that we can keep track, or do you think the time vs benefits are not worth to further improve the total time spent?

ggirotto:
I had another hypothesis about caching the Flutter setup, but since the time taken to setup Flutter env is of only ~50 seconds I don't think it will bring great benefits to the overall build time.

matuella:
Closing as we've reached a decent improvement with #98, although we can still look forward to more improvements if they appear plausible

PR #98, opened 2021-05-18
Improves iOS release workflow performance

ggirotto:
Relates to #94 improving the iOS release workflow performance. Previously we were building the iOS app twice. Now we take advantage from `flutter build ipa` to build the app once

### DIFF
--- a/ios/fastlane/Fastfile
+++ b/ios/fastlane/Fastfile
@@ -78,9 +78,6 @@ platform :ios do
     # Run Analyzer
     sh("flutter", "analyze", ".")
 
-    # Run flutter build with release mode
-    sh("flutter", "build", "ios", "--release", "--dart-define=ENV=PROD")
-
     # Increment the build number (use the latest build number for this version + 1)
     increment_build_number(
       build_number: latest_testflight_build_number(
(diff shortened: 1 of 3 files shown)


matuella (dismissed):
Really nice, didn't know about this flutter ipa build.

matuella:
@ggirotto  Can we merge it?

ios/fastlane/Fastfile
```diff
@@ -101,10 +101,8 @@ platform :ios do
       api_key: api_key,
       skip_waiting_for_build_processing: true,
       distribute_external: true,
-      changelog: changelog
+      changelog: changelog,
+      ipa: '../build/ios/ipa/Memo.ipa'
     )
-
-    # Clean local dSYM and IPA files
-    clean_build_artifacts
```
ggirotto, line 119: We don't need to clean artifacts anymore because `flutter build ipa` builds to a temporary directory, so the archive and the DSyms are not placed in the root directory anymore

ggirotto:
@matuella I cleaned the certificates, identifiers and provisions from my personal account, so I had to update Memo provisions as well. Please review my last two commits

matuella (dismissed):

PR #109, opened 2021-06-21
Adds `Firebase Crashlytics` dependency

ggirotto:


lib/data/repositories/analytics_monitor.dart
```diff
@@ -0,0 +1,47 @@
...
+    _setupCrashlytics();
+  }
+
+  static final AnalyticsMonitorImpl instance = AnalyticsMonitorImpl._();
+
+  Future<void> _setupCrashlytics() async {
+    if (kDebugMode) {
+      await FirebaseCrashlytics.instance.setCrashlyticsCollectionEnabled(false);
+    }
+  }
```
matuella, line 38: Not quite a fan of this approach. I would much rather receive an instance as an argument and this `FirebaseCrashlytics.instance.setCrashlyticsCollectionEnabled(false)` should be based on whether the `EnvMetadata` is dev or note. Also, it should be called before adding this same instance to this reporter.

I will add my overall suggestion on this PR comment.

README.md
```diff
@@ -44,6 +44,12 @@ If you have no idea how to install Flutter and run it locally, check this
 If you have Flutter setup locally, on the project's root folder, install pubspec dependencies by running
 `flutter pub get`.
 
+### Firebase dependencies
+
+Memo has Firebase dependencies that require a setup before the project can be run locally. The original 
+`GoogleServices-Info.plist` (iOS) and `google_service_account` (Android) are not checked in source control, which means
```
matuella, line 50: `google-services.json`

README.md
```diff
@@ -44,6 +44,12 @@ If you have no idea how to install Flutter and run it locally, check this
 If you have Flutter setup locally, on the project's root folder, install pubspec dependencies by running
 `flutter pub get`.
 
+### Firebase dependencies
+
+Memo has Firebase dependencies that require a setup before the project can be run locally. The original 
+`GoogleServices-Info.plist` (iOS) and `google_service_account` (Android) are not checked in source control, which means 
+that you must provide your own Firebase project google service files to run the app.
```
matuella, line 51: I would also add a link to people that may never had setup a firebase project, so they can start from somewhere:

`If you want to know how to setup your own Firebase project, check out the [FlutterFire docs](https://firebase.flutter.dev/docs/overview/)`

CHANGELOG.md
```diff
@@ -10,6 +10,14 @@ a beta or production release, they must be documented here).
 
 ## [Unreleased]
 
+## Added
+- `Firebase Crashlytics` dependency;
+- `AnalyticsMonitor`, which controls app errors logging and custom events;
+- Updated `release` workflow by adding dSYM upload to Crashlytics.
+
+## Updated
+- Bumped `strict` version.
```
matuella, line 19: I don't think this needs to be addressed

android/app/build.gradle
```diff
@@ -65,4 +66,9 @@ flutter {
 
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
+    implementation platform('com.google.firebase:firebase-bom:28.1.0')
+    implementation 'com.google.firebase:firebase-analytics-ktx'
```
matuella, line 70: What are these dependencies? Why they are needed?

android/app/build.gradle
```diff
@@ -65,4 +66,9 @@ flutter {
...
 }
+
+apply plugin: 'com.google.gms.google-services'
+apply plugin: 'com.google.firebase.crashlytics'
```
matuella, line 74:
```suggestion
```

android/app/build.gradle
```diff
@@ -23,6 +23,7 @@ if (flutterVersionName == null) {
 }
 
 apply plugin: 'com.android.application'
+apply plugin: 'com.google.gms.google-services'
```
matuella, line 26:
```suggestion
apply plugin: 'com.google.gms.google-services'
apply plugin: 'com.google.firebase.crashlytics'
```

matuella (changes requested):
I personally don't think it's useful to add an abstraction on top of the Crashlytics implementation, as this is appears to me like an over-engineered solution, it seems to me that it gives much more verbosity to the actual solution and makes things kind of confusing, as it's not clear where this is handled properly, because there is partial mention to this reporter class in `main`, `app_vm` and its own reporter class. I would much rather just make these changes to the `main` file and reset all other changes, like so:

```dart
  WidgetsFlutterBinding.ensureInitialized();

  final env = envMetadata();

  final crashlytics = FirebaseCrashlytics.instance;

  await Firebase.initializeApp();
  await crashlytics.setCrashlyticsCollectionEnabled(!env.isDev);
  FlutterError.onError = crashlytics.recordFlutterError;

  // Wraps `AppRoot` in a guarded zone where all errors are reported to `crashlytics.recordError`
  await runZonedGuarded<Future<void>>(
    () async {
      runApp(AppRoot(appVM));
    },
    crashlytics.recordError,
  );
```

With these changes, even though we are constrained to the `Crashlytics.instance` usage, it should be super easy to change for any different reporting tool, and it actually is called before ANYTHING else in the application, which is really important if we have some error in between the `AppVMImpl` instance and the actual `runApp` call.

android/app/build.gradle
ggirotto: Removed. They were suggest by Firebase docs but are not required at all

CHANGELOG.md
```diff
@@ -13,11 +13,11 @@ a beta or production release, they must be documented here).
 ## Added
 - `Firebase Crashlytics` dependency;
 - `AnalyticsMonitor`, which controls app errors logging and custom events;
-- Updated `release` workflow by adding dSYM upload to Crashlytics.
-
-## Updated
-- Bumped `strict` version.
+- `env` file that manages application environment state.
```
matuella, line 15: should be more specific, like`EnvMetadata`, `EnvMetadataImpl`

matuella (commented):
Is this `env` documented properly in the `ARCHITECTURE`?

ggirotto:
@matuella Added docs at `Environment` section from `ARCHITECTURE.md`

ARCHITECTURE.md
```diff
@@ -502,6 +502,10 @@ no more multiple `main.dart` files and such stuff. Simply run:
 If you are using `vscode` IDE, there is the [launch configuration files](.vscode/launch.json) for you to auto run and
 debug the application.
 
+The environment is managed by `EnvMetadata` implementation, located in [`env.dart` file](./lib/core/env.dart). It reads 
+`ENV` flag and provides metadata that are dependent from the current running environment - like database URLs for 
+example - to the rest of the application.
```
matuella, line 507:
```suggestion
During runtime, environment properties are provided through `EnvMetadata` implementation, located in [`env.dart` file](./lib/core/env.dart). It uses this same
`ENV` flag to provide the respective environment's metadata, such as server URLs.

If you are using `vscode` IDE, there is the [launch configuration files](.vscode/launch.json) for you to auto run and
debug the application.
```

matuella (approved):

PR #129, opened 2021-07-16
Updates Android & iOS release lanes to distribute beta build to external users

ggirotto:
Introduces a performance improvement to release workflow that may save around 5 minutes, according to the last release workflow logs:

CleanShot 2021-07-16 at 11 10 12@2x

It basically ignores the processing time from AppStoreConnect. It doesn't make sense to wait for it since we'll need to manually release it latter in the portal.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ a beta or production release, they must be documented here).
 
 ## [Unreleased]
 
+## Updated
+- Android & iOS Fastfiles to automatically distribute uploaded builds to external testers.
+
 ## [0.1.0] - 2021-07-16
 
 ## Added
--- a/android/fastlane/Fastfile
+++ b/android/fastlane/Fastfile
@@ -30,6 +30,7 @@ platform :android do
       skip_upload_images: true,
       skip_upload_screenshots: true,
       skip_upload_apk: true,
+      skip_upload_metadata: true,
       version_name: version_info["version_name"],
       version_code: android_get_version_code(),
       aab: "../build/app/outputs/bundle/release/app-release.aab",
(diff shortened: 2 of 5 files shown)


matuella (commented):
Didn't `distribute_external: true` automatically upload to all external testers?

ggirotto:
@matuella Yes, but I think we'll need to provide the correctly CHANGELOG to them anyway. I though about using `Changelog.txt` that I removed in this PR, but I don't think this is going to work unless we keep it updated as we updated the base CHANGELOG.md. What do you think?

matuella:
> @matuella Yes, but I think we'll need to provide the correctly CHANGELOG to them anyway. I though about using `Changelog.txt` that I removed in this PR, but I don't think this is going to work unless we keep it updated as we updated the base CHANGELOG.md. What do you think?

I think the `CHANGELOG.md` is too technical for Testflight testers. I think these are good to support, although we will have to update these alongside `CHANGELOG.md`.

On a side note, how Play Store deals with version changelogs (for testers)?

ggirotto:
We can have a similar structure in the Android target. [Fastlane Supply](https://docs.fastlane.tools/actions/supply/) supports CHANGELOG txt files as we are doing in iOS today. Do you think we should keep the CHANGELOG and automatically distribute new release to external testers or manually release these versions?

matuella:
> We can have a similar structure in the Android target. [Fastlane Supply](https://docs.fastlane.tools/actions/supply/) supports CHANGELOG txt files as we are doing in iOS today. Do you think we should keep the CHANGELOG and automatically distribute new release to external testers or manually release these versions?

I think automatic distributing for releases are nice, honestly - this means that store releases will still required a manual effort - as expected. Given that the work is pretty much summarizing the release in a .txt file, I don't see why not do it.

ggirotto:
@matuella Updated this PR to correctly distribute beta builds with their respective CHANGELOG to iOS and Android stores

matuella (approved):

PR #158, opened 2021-07-23
Fixes emojis overlap in `ExecutionTerminal`

ggirotto:


lib/application/pages/execution/execution_terminal.dart
```diff
@@ -158,6 +159,7 @@ class ExecutionTerminal extends HookWidget {
         child: Opacity(
           opacity: curvedController.value,
           child: _TerminalActions(
+            key: _terminalActionsKey,
```
matuella, line 162: shouldn't this be a constant `ValueKey` if you want it to not change between multiple builds? Also, why exactly adding a key here?

matuella (commented):
What do you think if using a `CustomMultiChildLayout` or `CustomScrollView`?

I'm not quite a fan of using `addPostFrameCallback` unless it's the only solution available. My reason is because it adds a layer of complexity that is really hard to grasp in many situations - as it's not a common usage.

I could be assuming something wrong here, but wouldn't this whole widget be rebuilt just because of this change? Because once the `WidgetsBinding` has flushed its contents (thus calling the first-frame callback), it would set the state of this hook to its new height. 

I quite like the smart solution, using the editor's padding to adjust its contents, but I just think that maybe using something like a custom scroll view may be the most performant and readable solution in this case.

matuella:
Closing this in favor of #187

PR #159, opened 2021-07-23
Fixes iOS and Android Splash Screens

### LINKED ISSUE #130: White screen when opening/loading the app
**Descreva o Bug/Describe the bug:**

There is a white page before displaying the `SplashPage`. If the white page takes too long (usually during the first application's initialization), the `SplashPage` doesn't even appear.

Possibly only iOS physical device, have to check if Android physical devices also have the same problem.

Have to check if the same occurs on emulators.

**Passos para Reproduzir/Steps to Reproduce:**

1. Open the app.

**Resultado Esperado/Expected Result:**

Always displaying the `SplashPage` until the application is fully loaded.

**Resultado Obtido/Actual Result:**

White page appears before the `SplashPage`.

**Versão do app/App Version:**

0.1.0, iOS 14.6 iPhone 6s

ggirotto:
Fixes #130


matuella (commented):
Is this a temporary solution? It doesn't seems to me that it's an assumed solution, given how there's no associated Flutter issue with this (and it's clearly an issue with the framework then).

Also, does this occurs on Android devices as well for you to update the Android solution?

matuella:
Maybe a regression issue? Loosely related: https://github.com/flutter/flutter/issues/37155

ggirotto:
Why do you believe this is not an assumed solution? Using the native splash screens avoids any possible Flutter problem and it's definitely the optimal solution, since it uses the native way of handling with splash screens instead relying on Flutter framework lifecycle.

I also updated Android because now the splash screen appears faster.

I don't think that #37155 is the same issue that we're facing, since the OP reports that the app freeze at the black screen, and we just have an unexpected delay (maybe Flutter framework initialize delay, not sure)

matuella:
> Why do you believe this is not an assumed solution? Using the native splash screens avoids any possible Flutter problem and it's definitely the optimal solution, since it uses the native way of handling with splash screens instead relying on Flutter framework lifecycle.

Well, if you're going that way, I could say the same for "implementing native UI screens is the optimal solution". If it's out of Flutter's scope of handling stuff, I would totally agree with you, such as handling camera usage, audio-related operations, etcetera. - we could even possibly create a plugin for such thing -, but again, I don't believe this case. A splash screen is something that's not required, but it's a standard that all mobile applications, and it's definitely something that we should expect from Flutter's framework - and after all, it's just a screen.

What I feel from this is simply a limitation - or some bug - with Flutter, thus why it doesn't feel to me like a solution, more like a workaround.

matuella (commented):
> Why do you believe this is not an assumed solution? Using the native splash screens avoids any possible Flutter problem and it's definitely the optimal solution, since it uses the native way of handling with splash screens instead relying on Flutter framework lifecycle.

Well, if you're going that way, I could say the same for "implementing native UI screens is the optimal solution". If it's out of Flutter's scope of handling stuff, I would totally agree with you, such as handling camera usage, audio-related operations, etcetera. - we could even possibly create a plugin for such thing -, but again, I don't believe this case. A splash screen is something that's not required, but it's a standard that all mobile applications, and it's definitely something that we should expect from Flutter's framework - and after all, it's just a screen.

What I feel from this is simply a limitation - or some bug - with Flutter, thus why it doesn't feel to me like a solution, more like a workaround.

ggirotto:
I disagree this is a workaround, since it's suggested even in the [Flutter docs](https://flutter.dev/docs/development/ui/advanced/splash-screen). They state the same that I believe:

> They (SplashScreen) set the stage for your application, while allowing time for the app engine to load and your app to initialize.

If there is the possibility of Flutter engine delays it initialization, we cannot rely on it Flutter to provide a solution to the problem. They also state:

> Each Flutter experience in an app requires a few moments to initialize the Dart isolate that runs the code. This means that a user momentarily sees a blank screen until Flutter renders its first frame.

Which is the exactly description about this issue. They propose a ["Flutter" solution](https://flutter.dev/docs/development/ui/advanced/splash-screen#the-flutter-splash-screen) that uses a Android View, which I personally believe that it's worst from the platform-related solution proposed by this PR

matuella (approved):
> I disagree this is a workaround, since it's suggested even in the Flutter docs.

Indeed, this is what the Flutter's team recommends to do - I was completely wrong. Interesting article, haven't seen this one when I read the docs last time!

Nonetheless, I still feel like this is something that we could think about improving, although I think it won't be trivial. My only argument to why this can lead to future problems is that a future change to the `SplashPage` is not guaranteed to be equal in both native platforms, that's my only concern with 3 different ecosystems requiring the exact same interface, that may change over time. Even though this is the ideal solution, given the cited article, I suggest it to have a little succinct section in our `ARCHITECTURE`, as a reference for future changes. This sections is a nitpick, it's up to you to add it.

PR #165, opened 2021-07-24
Worfklows performance improvements

ggirotto:
Relates to https://github.com/olmps/memo/issues/94

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -37,9 +37,8 @@ jobs:
           flutter-version: ${{ env.flutter_version }}
 
       # Run pre-requisites to build both applications (iOS and Android)
-      - name: Clean, Analyze and run tests
+      - name: Analyze and run tests
         run: |
-          flutter clean
           flutter analyze .
           flutter test test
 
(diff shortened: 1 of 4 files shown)


.github/workflows/pull-request.yml
```diff
@@ -23,16 +23,22 @@ jobs:
...
       - name: Setup Flutter
         uses: subosito/flutter-action@v1
         with:
           flutter-version: ${{ env.flutter_version }}
 
       # Run pre-requisites to build both applications (iOS and Android)
-      - name: Clean, Analyze and run tests
+      - name: Analyze and run tests
         run: |
-          flutter clean
```
ggirotto, line 35: There's no need to run `flutter clean` since we just did a fresh install of it.

.github/workflows/release.yml
```diff
@@ -48,6 +55,12 @@ jobs:
           known_hosts: ${{ secrets.KNOWN_HOSTS }}
           if_key_exists: replace
 
+      # Run pre-requisites to build both applications (iOS and Android)
+      - name: Analyze and run tests
+        run: |
+          flutter analyze .
+          flutter test test
```
ggirotto, line 62: Moves `analyze` and `tests` from both targets Fastfiles to the workflow. The analyze and tests result don't change based on the platform, so it was a unnecessary duplicate work

matuella (approved):
Really nice improvements

PR #166, opened 2021-07-26
Analytics 

ggirotto:


PRIVACY_POLICY.md
```diff
@@ -0,0 +1,28 @@
+Your privacy is critically important to us. This document states what data we collect during Memo's usage and why we
```
matuella, line 1: Wouldn't this file conflict directly with terms that is being shared in the application's settings? Overall, I think this is a really good condensed version of the privacy policy (although I'd improve REALLY minor grammar stuff), so we should evaluate how this overlap with the one that is seen in the app.

matuella: Also, build failed

ggirotto: @matuella 

I was thinking about renaming this file to be `ANALYTICS.md` or something unrelated to privacy police, so we can explain in a less "juridic" language what data we collect and how we use it. This way we can keep our current privacy policy, which is more "generic" and keep this as a separate file to be a more informal and less verbose explanation about the data we collect and how we use it.

Or we can merge both of them, but this option would take a little more effort to achieve, since we've to think in other aspects than just collected information.

What do you think?

matuella: I'm not so sure, never seen `ANALYTICS.md` as a standard. I think we should standardize all of it and include in this repository - including the one that we are sharing for "all memo stuff", that generic that includes basically everything. For now, I would remove this from the PR and we should think on how to merge them both, although surely in a new PR.

matuella: I'm not so sure, never seen `ANALYTICS.md` as a standard. I think we should standardize all of it and include in this repository - including the one that we are sharing for "all memo stuff", that generic that includes basically everything. For now, I would remove this from the PR and we should think on how to merge them both, although surely in a new PR.

matuella (approved):

PR #167, opened 2021-07-27
Adds sksl to reduce animations stutter

ggirotto:


android/app/build.gradle
```diff
@@ -39,8 +39,9 @@ android {
 
     defaultConfig {
         applicationId "com.olmps.memoClient"
-        minSdkVersion 16
+        minSdkVersion 17
         targetSdkVersion 30
+        multiDexEnabled true
```
ggirotto, line 44: `quill_version` now depends on `youtube_player_flutter` that depends on `flutter_inapp_webview` that [requires](https://github.com/pichillilorenzo/flutter_inappwebview/blob/f06bcdf695847a0187a31ab607b2bfe236fb0057/android/build.gradle#L28) min sdk version 17 and `multiDexEnabled` enabled

lib/application/pages/details/collection_details_page.dart
```diff
@@ -4,10 +4,10 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:layoutr/common_layout.dart';
```
ggirotto, line 4: Sorted the imports alphabetically (`flutter analyze` complains if not)

matuella: Do you know if this is linteable? It seems that it should, if flutter is complaining.

ios/Flutter/AppFrameworkInfo.plist
```diff
@@ -21,6 +21,6 @@
   <key>CFBundleVersion</key>
   <string>1.0</string>
   <key>MinimumOSVersion</key>
-  <string>8.0</string>
+  <string>9.0</string>
```
matuella, line 24: Why the iOS minimum version was also bumped?

matuella (commented):
Should both `.json` files be committed? Shouldn't these shaders be generated whenever any Dart code changes?

ios/Flutter/AppFrameworkInfo.plist
ggirotto: my bad, it automatically bumped when I updated flutter to the beta version (don't know exactly why, seems related with https://github.com/flutter/flutter/issues/58200)

ggirotto: I don't know why Flutter is automatically bumping this version to 9 every time I build the iOS app:

CleanShot 2021-07-28 at 17 15 56

There's nothing about the minimum version change in the [Breaking Changes](https://flutter.dev/docs/release/breaking-changes) guide, although they state in [Build and release an iOS app
](https://flutter.dev/docs/deployment/ios#review-xcode-project-settings) that the minimum supported version is iOS 9.0 (we also have deployment target set to 9.0 in the PBX)

CleanShot 2021-07-28 at 17 19 07
CleanShot 2021-07-28 at 17 18 58

TLDR: I don't know why Flutter automatically bumped, but the version 9 seems to be the correct one

ggirotto:
> Should both .json files be committed? Shouldn't these shaders be generated whenever any Dart code changes?

Ideally we should have integration tests that run the app and re-generate these JSONs every time the code changes, as suggested in [Flutter article](https://flutter.dev/docs/perf/rendering/shader#what-do-we-mean-by-first-run). Using these pre-compiled JSONs is a short term solution that may be replaced in the future with these integration tests (or even with a Flutter solution)

lib/application/pages/details/collection_details_page.dart
ggirotto: This is being catch by the linter:

CleanShot 2021-07-28 at 17 26 14

I'm just thinking why Flutter wasn't complaining before about it, but I have no idea why

matuella:
> Ideally we should have integration tests that run the app and re-generate these JSONs every time the code changes, as suggested in Flutter article. Using these pre-compiled JSONs is a short term solution that may be replaced in the future with these integration tests (or even with a Flutter solution)

This seems relatively easy to do, and I'm not really inclined to add a version that require a manual change every time a new version is required - and we are aware on how to improve it, meaning, already creating a high priority issue to it. I'm open to do this myself if you want to.

ggirotto:
> This seems relatively easy to do, and I'm not really inclined to add a version that require a manual change every time a new version is required - and we are aware on how to improve it, meaning, already creating a high priority issue to it. I'm open to do this myself if you want to.

I'm afraid that these integration tests will be very tricky because we'll basically need integration tests for the entire app animations (scroll the main collections list, open a deck details, close deck details, start a deck, answer a question, etc...).

Basically we need to cover all actions that we've inside the app today. Although I've never implemented integration tests, it seems at least toilsome to do so, and we have other priorities (server) that have higher priority right now.

Do you have something in mind like a MVP for these tests? I think that we can start using a manual approach like the one proposed by this PR and create a project to start implementing these tests and lather remove these animation JSONs

matuella (approved):

PR #156, opened 2021-07-23
Pull request Workflow Fix

ggirotto:
Fixes forked PR not being able to run `pull_request` workflow due to the usage of Github Secrets

PR #168, opened 2021-07-27
Fixes Android debug build

ggirotto:
https://github.com/olmps/memo/pull/156 introduced an issue that breaks VSCode build in Android devices (physical and simulators), because Firebase service-account is not used

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -23,7 +23,7 @@ if (flutterVersionName == null) {
 }
 
 apply plugin: 'com.android.application'
-if (System.getenv("IS_SIGNING_ENABLED") == "true") {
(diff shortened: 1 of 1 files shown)


android/app/build.gradle
```diff
@@ -23,7 +23,7 @@ if (flutterVersionName == null) {
 }
 
 apply plugin: 'com.android.application'
-if (System.getenv("IS_SIGNING_ENABLED") == "true") {
+if (System.getenv("IS_SIGNING_ENABLED") == "true" || System.getenv("IS_SIGNING_ENABLED") == null) {
```
ggirotto, line 26: `IS_SIGNING_ENABLED` will be `null` when building locally. We must use firebase plugins when building locally, otherwise the app will break

matuella (dismissed):

PR #171, opened 2021-07-29
Redesigned empty state

ggirotto:


lib/application/constants/images.dart
```diff
@@ -17,6 +17,7 @@ enum ImageKey {
   trendingUpArrow,
 
   // illustrations
+  folderIllustration,
```
ggirotto, line 20: This is the same as `folder` asset, but since `folder` was exported to fit in the tab, it hasn't enough dimensions to be used in the empty state (it gets extremely low quality if stretched to the desired dimensions)

lib/application/pages/home/collections/collections_page.dart
```diff
@@ -91,3 +95,51 @@ class CollectionsContents extends HookWidget {
...
+          strings.collectionsEmptyMessageSegment(currentSegment),
+          style: textTheme.bodyText2?.copyWith(color: theme.neutralSwatch.shade400),
+          textAlign: TextAlign.center,
+        ),
+        context.verticalBox(Spacing.xLarge),
+        ElevatedButton(
+          onPressed: () {
+            final oppositeSegment =
+                currentSegment == CollectionsSegment.explore ? CollectionsSegment.review : CollectionsSegment.explore;
+            onUpdateSegment(oppositeSegment);
```
ggirotto, line 137: Not sure if this is the best solution. Open to suggestions

lib/application/constants/images.dart
matuella: Then maybe `folderBig`?

matuella: Naming it as "illustration" makes others seem to not adhere to this nomenclature

lib/application/pages/home/collections/collections_page.dart
```diff
@@ -91,3 +94,51 @@ class CollectionsContents extends HookWidget {
     return widget.withSymmetricalPadding(context, horizontal: Spacing.medium);
   }
 }
+
+/// [CollectionsPage] empty state, given the current [collectionsVM] state.
```
matuella, line 98:
```suggestion
```

Seems like an irrelevant doc for me, which is incorrectly referring to be related to [collectionsVM] - [_CollectionsContents] is.

lib/application/pages/home/collections/collections_page.dart
```diff
@@ -91,3 +94,51 @@ class CollectionsContents extends HookWidget {
...
+    final theme = useTheme();
+
+    return Column(
+      mainAxisAlignment: MainAxisAlignment.center,
+      children: [
+        Image.asset(
+          images.folderIllustrationAsset,
+          height: dimens.collectionsEmptyStateImageSize,
+          width: dimens.collectionsEmptyStateImageSize,
+          fit: BoxFit.fill,
```
matuella, line 117: fill removes the aspect ratio. You probably want it to be contained

lib/application/constants/dimensions.dart
```diff
@@ -23,6 +23,7 @@ const double smallIconSize = 16;
 // CollectionsPage
 //
 const double collectionsLinearProgressLineWidth = 12;
+const double collectionsEmptyStateImageSize = 96;
```
matuella, line 26:
```suggestion
const double collectionsEmptyStateSize = 96;
```

lib/application/pages/home/collections/collections_page.dart
matuella: I would remove nested check here, as it can lead to future problems if new tabs are added/updated.

My suggestion is to delegate its responsibility to the caller, something like:

```dart
//...
    child: _CollectionsContents(
      onSegmentSwapRequested: () {
   final nextIndex = segment.index + 1 % availableSegments.length;
   collectionsTabController.animateTo(nextIndex);
},
    ),
//

//...
class _CollectionsContents extends HookWidget {
  const _CollectionsContents({this.onSegmentSwapRequested});
  
  /// {@macro...}
  final VoidCallback onSegmentSwapRequested;
//...
}

//...
class _CollectionsEmptyState extends HookWidget {
  const _CollectionsEmptyState({required this.onSegmentSwapRequested, required this.title, required this.description});
  
  /// {@template}
  /// An empty-state request for a segment swap.
  /// ....
  final VoidCallback onSegmentSwapRequested;
 //...
```

matuella (changes requested):

matuella (approved):

PR #173, opened 2021-07-29
Replace `AlertDialog` by `showModalBottomSheet`

ggirotto:
It also adds a new color swatch named `negative`

### DIFF
--- a/lib/application/constants/strings.dart
+++ b/lib/application/constants/strings.dart
@@ -75,7 +75,7 @@ const executionNext = 'Próxima';
 const executionCheckAnswer = 'Ver resposta';
 
 const executionDiscard = 'Descartar';
-const executionBackToStudy = 'Voltar ao estudo';
+const executionBackToStudy = 'Voltar ao Estudo';
 const executionDiscardStudy = 'Descartar Estudo';
 const executionDiscardStudyDescription =
     'Se você encerrar seu estudo, perderá todo o seu progresso.\nDeseja prosseguir?';
--- a/lib/application/constants/strings.dart
+++ b/lib/application/constants/strings.dart
@@ -74,6 +74,8 @@ const executionAnswer = 'Resposta';
 const executionNext = 'Próxima';
 const executionCheckAnswer = 'Ver resposta';
 
+const executionDiscard = 'Descartar';
+const executionBackToStudy = 'Voltar ao estudo';
 const executionDiscardStudy = 'Descartar Estudo';
 const executionDiscardStudyDescription =
     'Se você encerrar seu estudo, perderá todo o seu progresso.\nDeseja prosseguir?';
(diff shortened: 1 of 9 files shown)


lib/application/theme/memo_theme_colors.dart
```diff
@@ -46,6 +46,25 @@ MaterialColor buildClassicSecondarySwatch() {
   );
 }
 
+MaterialColor buildClassicNegativeSwatch() {
```
ggirotto, line 49: I also like `danger` or `destructive`. Open to suggestions

matuella: I do prefer destructive - totally personal preference, would be good with negative as well.

Danger is too specific

lib/application/widgets/theme/negative_button.dart
```diff
@@ -0,0 +1,22 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_hooks/flutter_hooks.dart';
+import 'package:memo/application/theme/theme_controller.dart';
+
+/// Negative alternative to [ElevatedButton].
+///
+/// Used in scenarios where the user action may be destructive
```
matuella, line 7:
```suggestion
/// Destructive alternative to [ElevatedButton].
///
/// Used in scenarios that emphasizes an action with impactful consequences.
```

lib/application/widgets/theme/negative_button.dart
```diff
@@ -0,0 +1,22 @@
...
+class NegativeButton extends HookWidget {
+  const NegativeButton({required this.onPressed, required this.child, Key? key}) : super(key: key);
```
matuella, line 9: Just like I suggested before, I think destructive would be better - but really personal, I'd say negative is good as well.

My only rationale with preferring the former is that destructive is not necessarily a negative behavior - meaning that destroying something is completely acceptable, although requires enough emphasis.

lib/application/constants/strings.dart
```diff
@@ -65,6 +65,8 @@ const executionAnswer = 'Resposta';
 const executionNext = 'Próxima';
 const executionCheckAnswer = 'Ver resposta';
 
+const executionDiscard = 'Descartar';
+const executionBackToStudy = 'Voltar ao estudo';
```
matuella, line 69:
```suggestion
const executionBackToStudy = 'Voltar ao Estudo';
```

Following the same as the destructive action text

matuella (changes requested):

lib/application/theme/memo_theme_colors.dart
ggirotto: Updated do destructive

lib/application/widgets/theme/negative_button.dart
ggirotto: Updated do destructive

matuella (approved):

Issue #172, opened 2021-07-29
Missing Safe Area during Collection execution


ggirotto:
Similar to https://github.com/olmps/memo/issues/95, but related to execution screen

![CleanShot 2021-07-29 at 09 17 49@2x](https://user-images.githubusercontent.com/11745745/127490371-889d7dd2-f6e8-4dc5-9927-d447bbe9a4d0.png)

PR #174, opened 2021-07-30
Fixes SafeArea in `ExecutionTerminal`

ggirotto:
Fixes https://github.com/olmps/memo/issues/172

### DIFF
--- a/lib/application/pages/execution/execution_terminal.dart
+++ b/lib/application/pages/execution/execution_terminal.dart
@@ -106,25 +106,27 @@ class ExecutionTerminal extends HookWidget {
       child: Text(actionText.toUpperCase()),
     );
 
-    return Container(
-      clipBehavior: Clip.hardEdge,
-      decoration: BoxDecoration(
-        color: Colors.transparent,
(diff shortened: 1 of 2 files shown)


lib/application/pages/execution/execution_terminal.dart
```diff
@@ -106,25 +106,27 @@ class ExecutionTerminal extends HookWidget {
...
 
-    return Container(
-      clipBehavior: Clip.hardEdge,
-      decoration: BoxDecoration(
-        color: Colors.transparent,
-        borderRadius: dimens.executionsTerminalBorderRadius,
-        border: Border.all(
-          color: borderColor,
-          width: dimens.executionsTerminalBorderWidth,
+    return SafeArea(
```
matuella, line 109: I don't think this belongs to this particular widget. Instead, should be used by the page that uses this widget, because we are not sure where it would be used in the future. I would say somewhere [around here](https://github.com/olmps/memo/blob/main/lib/application/pages/execution/collection_execution_page.dart#L32).

matuella (approved):

PR #178, opened 2021-08-02
Server firebase files

matuella:
Adds core firebase files.

Documentation on these (including scripts and folder structure) will be added in a future PR.


firebase/firebase.json
```diff
@@ -0,0 +1,35 @@
...
+  "functions": {
+    "source": "src/functions",
+    "predeploy": [
+      "npm --prefix \"$PROJECT_DIR\" run functions-build"
+    ]
+  },
+  "emulators": {
+    "storage": {
+      "port": 9199
+    },
```
ggirotto, line 18: Firebase Storage emulator was launched recently and still have a lot of bugs. I would remove it while it's not stable

firebase/firebase.json
```diff
@@ -0,0 +1,35 @@
...
+    ]
+  },
+  "functions": {
+    "source": "src/functions",
+    "predeploy": [
+      "npm --prefix \"$PROJECT_DIR\" run functions-build"
+    ]
+  },
+  "emulators": {
+    "storage": {
```
ggirotto, line 16: Firebase Storage emulator was launched recently and still have a lot of bugs. I would remove it while it's not stable

firebase/firebase.json
```diff
@@ -0,0 +1,35 @@
...
+      "port": 9199
+    },
+    "auth": {
+      "port": 9099
+    },
+    "pubsub": {
+      "port": 8085
+    },
```
ggirotto, line 24: Why adding pubsub if we don't use it?

firebase/firebase.json
ggirotto: Also, we don't use Storage right?

matuella: I've checked their docs and it seems pretty good all around (even `@firebase/rules-unit-testing` received a significant update with these recent changes). Anyways, these are just standard ports, having these here or not won't make any difference - anyone can start a emulator locally if they wanted to, although all scripts in `package.json` only emulate the specific services that we are planning to use, which are `firestore` and `functions`

firebase/firebase.json
matuella: The comment above address your question as well.

ggirotto (approved):

matuella:
All updates made here will be merged in the last PR (#149) of this nested chain of PRs.

PR #194, opened 2021-09-27
Add repository root docs

matuella:

PR #195, opened 2021-09-27
Update Flutter docs

matuella:


flutter/ARCHITECTURE.md
```diff
@@ -37,43 +34,30 @@ Table of contents
...
+# Architecture
 
-  - [`launch.json`](.vscode/launch.json) is where all pre-configured command-line scripts are at, such as running a
-  debug dev environment;
-  - [`settings.json`](.vscode/settings.json) is responsible for the editor configurations, such as line-length, rules
-  and auto-format on save.
+## `.vscode/`
 
-### Useful vscode extensions
+There is not much to add on top of the [General ARCHITECTURE.md .vscode section](//TODO), although for this Flutter
```
ggirotto, line 51: Can't it be referenced from [here](https://github.com/olmps/memo/pull/194/files#diff-8f6366fd8e7a5fa30f7f05879999195b7cf5fa1e3d157822eb37f0e91d226cfdR58) ? This branch can point to that PR base branch.

flutter/ARCHITECTURE.md
```diff
@@ -122,16 +106,13 @@ making things slower just to follow some principles that don't necessarily apply
...
 
-One extra thing: this is heavily influenced by a bunch personal opinion and experiences in some production projects
-that the team has worked on. This project's external dependencies will keep changing as the time goes on, Flutter will
-also keep evolving, and we have to adapt in a way to maintain consistency, integrity and scalability of our solution.
-So, it's probable that there is (or will be) better ways to achieve the same goals/objectives, and for this, we look
-into your help to make this project's architecture continuously provide a good developer experience to add new features,
-update old ones and keep those nasty bugs away.
+Please, keep the [General ARCHITECTURE.md](../ARCHITECTURE.md#one-last-thing) introduction in mind - there are a ton of
+personal opinions and experiences here. This means that there **will be better ways to build what we've done**, and it's
+what we are aiming for: your help to improve this architecture over the time.
```
ggirotto, line 111: NIT: Same as https://github.com/olmps/memo/pull/194/files#r716847997

flutter/ARCHITECTURE.md
```diff
@@ -484,32 +461,15 @@ signature is: `showExceptionSnackBar(BaseException exception)`.
...
-versus expected version differs. You can find more about this in the `VersionServices` implementation file.
+If you change the asset, you'll have to run this tool again to update the required files:
+`flutter pub run flutter_launcher_icons:main`
 
 ## Environment
 
-For different types of build environments, we don't use the common _flavors_, iOS schemas and all of that painful setup,
-due to the fact that, since Flutter `1.17`, we can now use command arguments to inject any variable in our application -
+For different types of build environments, we **don't** use flavors, iOS schemas and all of that painful setup, due to
+the fact that, since Flutter `1.17`, we can now use command arguments to inject any variable in our application -
```
ggirotto, line 472: This is going to change when adding Firebase, isn't it? If using different containers to split development and production environments in Firebase, we'll need to use flutter flavors because we'll have different iOS bundle identifier and Android package identifiers for development and production respectively

flutter/ARCHITECTURE.md
```diff
@@ -521,56 +481,46 @@ such as server URLs.
...
+other phases that are required to deploy Memo.
+
+Under the hood, all of the heavy work is done by [Fastlane](https://fastlane.tools/). In simple terms, this tool
+automates the build and deploy to the platform's environment. After the deploy is completed, it bumps the build version
+from the targets. We currently have Fastlane running for [iOS](ios/fastlane/Fastfile) and
+[Android](android/fastlane/Fastfile) targets.
+
+> There's a catch to both iOS and Android build-versioning: the Android's build number is indefinitely incremented,
+> since it's not possible to reset an Android project build number, while iOS resets its build number every time a new 
+> version (major, minor or patch, doesn't matter) is bumped.
```
ggirotto, line 521: I don't think this comment aggregates in the document, since this is basically knowledge about Android versioning

flutter/README.md
```diff
@@ -1 +1,26 @@
...
+## Setup
+
+1. If you have no idea how to install Flutter and run it locally, check this
+[_Get started_](https://flutter.dev/docs/get-started/install).
+2. If you have Flutter setup locally, install pubspec dependencies by running `flutter pub get` in `flutter/` root
+folder.
+3. Setup Firebase dependencies. More information on this in [ARCHITECTURE#firebase](flutter/ARCHITECTURE.md#firebase).
+
+
+... and you're good to go - just run in any physical device, or a simulator/emulator.
```
ggirotto, line 17: I don't know if physical device will work for iOS without provision profiles setup

flutter/ARCHITECTURE.md
matuella: Your link leads to no specific comment for some reason. Anyways, I've answered them there.

flutter/ARCHITECTURE.md
matuella: I don't think so, we won't have development/production environments. We will possible have only local and production, and this could possible still coexist with using only CLI args

flutter/ARCHITECTURE.md
matuella: It gives context to those who don't know about such problem, and that's why this doc exist, to explain why - as it's not a technical code documentation, not even mentioning that assuming most of the developers know about this.

This is a document to explain the decisions, so that we don't forget and come back to discuss things like, in this case, "why don't we use the same build number for both platforms and improve consistency between them?".

flutter/README.md
matuella: Why not? It can simply ignore the provisions and change to its own developer account

flutter/ARCHITECTURE.md
ggirotto: But we can make some definitions today. Using Firebase Emulators for development without a development container in Firebase console has known impacts today, and it's necessary to decide if we are going to face them or not. I'll list some of them below

- [Firebase Emulator limitations](https://firebase.google.com/docs/emulator-suite/connect_firestore#limitations)
- Storage emulator is barely unuseful because it was recently launched and have tons of [bugs and errors](https://github.com/firebase/firebase-tools/issues?q=is%3Aopen+is%3Aissue+label%3A%22emulators%3A+storage%22)

Just to be clear: I'm not saying about the docs, I'm making this statements because their answer may change the server-side development in the short-term. I'm ok with updating the docs later (or not if we are going to use Firebase Emulators only), but we have to define if we need a development container or not.

flutter/README.md
ggirotto: But if the user needs an extra step to launch, shouldn't it be noted here?

flutter/ARCHITECTURE.md
ggirotto: I was trying to reference this comment: https://github.com/olmps/memo/pull/194#discussion_r716809096, but that's just me personal opinion, I have no arguments why it would be better than yours writing style, so let's move on

flutter/ARCHITECTURE.md
ggirotto: I was trying to say is that this is not a problem, and we not even changed that. It is working as it is supposed to be.

As a comparison example, I think it's worth it to mention when we remove `version` from `pubspec.yaml` when both iOS and Android targets have different versions, or like when we modify the Flutter generated `Podfile` to add some optimization or stuff like that, because in these situations we changed the default way of how thinks worked due to some project specific requirements.

This versioning docs just states what happens by default: iOS resets in every new version while Android versions the build incrementally. This happens even in native projects. We haven't changed that, that's why I think that this docs are not aggregating anything, they are just for educational purposes.

flutter/ARCHITECTURE.md
matuella: > Firebase Emulator limitations

I'm fine with these today. If we do need these indexes, we can think about adding a whole environment setup in the future.

> Storage emulator is barely useful* because it was recently launched and have tons of bugs and errors

Not used today, and, if necessary, the same logic above applies to this.

flutter/ARCHITECTURE.md
matuella: Yeah, I got your point, but like I've said, stating the obvious may not be obvious for everyone. This is a contextual knowledge that may teach something to people that didn't knew about that.

Think about some iOS developer that never adventured in the Android world - and vice-versa. They could be experienced developers, although they lack this definition, and adding here was pertinent enough to make them understand why these platform's version differ from each other and how we version the application, given this condition.

matuella:

flutter/README.md
matuella: I think this goes a little bit off of this explanation, as it's related to the setup of this project in particular, but I do get your point. I'm open to change it, although my opinion is to not add it. What's your suggestion on this one?

flutter/ARCHITECTURE.md
ggirotto: I get it, but where is the limit between contextual knowledge and we actually documented extra-stuff that is specific context knowledge? In my point of view, we should assume some knowledge from the public when writing the `ARCHITECTURE.md`, and highlight the decisions that may differ from other Flutter projects and specific changes that we've made to the ecosystem, like updating the Podfile or creating different iOS schemas, as examples. If not, we may assume that the public knows barely nothing about the Flutter ecosystem and explain everything. I know that this is not your point of view, I'm just making an absurd example to question: where the boundary lives? Again, in my opinion, the boundary are decisions that may be different in other Flutter projects and ecosystem changes that we made, and versioning is not one of them IMO.

I'm ok with the decision of keeping it, I just personally think it doesn't aggregates anything, but I may be wrong here. I leave this decision to you since I'm ok with both scenarios, just wanted to state what I think that should be in this document and what I think it shouldn't.

flutter/README.md
ggirotto: I would add an extra step:

`4. [Optional] To run in physical iOS devices, it's necessary to setup the provision profiles accordingly. You may change them by opening the XCode project and updating the Development Team in the target Signing & Capabilities to your personal developer account.`

What do you think?

matuella: I think it's pretty verbose, but I will try to make them more succinct while keeping the idea

flutter/ARCHITECTURE.md
matuella: > we should assume some knowledge from the public when writing the `ARCHITECTURE.md`

Definitely, but this is far too much to expect from anyone. This assumes that someone has knowledge on the deploy process. There may be people who are not directly in charge of this process in their respective projects, or even they haven't had any production projects - they simply never bothered to go through it. People may also come from Flutter/mobile from other non-traditional backgrounds, such as desktop, IoT, etcetera.

> where the boundary lives?

In all honesty, there is no such thing. This is too subjective and, even if we delimit the edges of this, like "when a developer that has worked with X has probably faced that", it'll be blurry anyways.

> it doesn't aggregates anything,

It does, but not for you, as you already know about this.

matuella: The exact same logic occurs with your https://github.com/olmps/memo/pull/195#discussion_r716861246 suggestion.

One should know that this is something to expect if you have knowledge on the Apple development ecosystem - but it's not true for everyone, because just like I've above, people come from different backgrounds and have different experiences.

ggirotto (approved):

PR #202, opened 2021-09-27
Add `FirestoreGateway`

matuella:
I accidentally added `createSinonStub` in this PR, so please, bear with me on this one, it'll make sense in the next chained PRs for this gateway's tests.

### DIFF
--- a/firebase/src/core/data/gateways/firestore-collection-name.ts
+++ b/firebase/src/core/data/gateways/firestore-collection-name.ts
@@ -1,0 +1,22 @@
+/** Typed firestore collections. */
+export type FirestoreCollectionName =
+  | PublicCollectionName
+  | PublicSubCollectionName
+  | UserCollectionName
+  | UserSubCollectionName;
+
+/** Public collections available to all authenticated users. */
+type PublicCollectionName = "public_collections" | "public_collection_categories" | "public_memos";
+
+/** Public subcollections available to all authenticated users. */
+type PublicSubCollectionName = `public_collections/${string}/public_memos`;
+
+type UserCollectionName = "users";
+
+/** User-owned subcollections available to all respective authenticated users. */
+type UserSubCollectionName =
+  | `${UserCollectionName}/${string}/collections`
+  | `${UserCollectionName}/${string}/collections/${string}/memos`
+  | `${UserCollectionName}/${string}/collection_categories`
+  | `${UserCollectionName}/${string}/collection_executions`
+  | `${UserCollectionName}/${string}/collection_executions/${string}/memo_executions`;
--- a/firebase/src/core/data/gateways/firestore-collection-name.ts
+++ b/firebase/src/core/data/gateways/firestore-collection-name.ts
@@ -1,22 +1,19 @@
-/** Typed firestore collections. */
-export type FirestoreCollectionName =
-  | PublicCollectionName
-  | PublicSubCollectionName
-  | UserCollectionName
-  | UserSubCollectionName;
+/** Typed firestore collections paths. */
+export type FirestorePaths = PublicPaths | UsersPaths;
+
+/** Raw string that represent any collection's document id. */
+type Id = string;
 
 /** Public collections available to all authenticated users. */
-type PublicCollectionName = "public_collections" | "public_collection_categories" | "public_memos";
+type PublicPaths = PublicCollectionsPaths | "collection_categories";
 
-/** Public subcollections available to all authenticated users. */
-type PublicSubCollectionName = `public_collections/${string}/public_memos`;
+/** Collection-related paths (the entity, not the Firestore one) paths, available to all authenticated users. */
+type PublicCollectionsPaths = "collections" | `collections/${Id}/memos`;
 
-type UserCollectionName = "users";
-
-/** User-owned subcollections available to all respective authenticated users. */
-type UserSubCollectionName =
-  | `${UserCollectionName}/${string}/collections`
-  | `${UserCollectionName}/${string}/collections/${string}/memos`
-  | `${UserCollectionName}/${string}/collection_categories`
-  | `${UserCollectionName}/${string}/collection_executions`
-  | `${UserCollectionName}/${string}/collection_executions/${string}/memo_executions`;
+/** User-related collection paths, which are available to the respective authenticated users. */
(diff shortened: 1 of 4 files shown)


firebase/src/core/data/gateways/firestore-collection-name.ts
```diff
@@ -0,0 +1,22 @@
+/** Typed firestore collections. */
```
ggirotto, line 1: Nice usage of types

firebase/src/core/data/gateways/firestore-gateway.ts
```diff
@@ -0,0 +1,235 @@
...
+/** Wraps {@link Firestore} operations, allowing transactions and typed collections. */
+export class FirestoreGateway<Path extends string> {
+  readonly #firestore: Firestore;
+
+  constructor(firestore: Firestore) {
+    this.#firestore = firestore;
+  }
+
+  /** Current transaction, if there is an ongoing {@link FirestoreGateway.runTransaction}. */
+  #transaction: Transaction | null = null;
```
ggirotto, line 39: I'm not sure how the use-cases will be instantiated, but using a Transaction here imply that only a single instance of this gateway will be used across the function container. If that it's expected, let's move on.

firebase/src/core/data/gateways/firestore-collection-name.ts
```diff
@@ -0,0 +1,22 @@
...
+export type FirestoreCollectionName =
+  | PublicCollectionName
+  | PublicSubCollectionName
+  | UserCollectionName
+  | UserSubCollectionName;
+
+/** Public collections available to all authenticated users. */
+type PublicCollectionName = "public_collections" | "public_collection_categories" | "public_memos";
```
ggirotto, line 9: is the `public` suffix necessary since they'll be placed in the root and the private ones under `users/`? It's seems intuitive IMO

firebase/src/core/data/gateways/firestore-collection-name.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/** Public subcollections available to all authenticated users. */
+type PublicSubCollectionName = `public_collections/${string}/public_memos`;
```
ggirotto, line 12: Even this one, `public_memos` that are placed in an already "public" collection (as its name suggests)

firebase/src/core/data/gateways/firestore-collection-name.ts
matuella: Yeah, we should be able to differ from collection groups that belong to users versus public ones. If we don't have this distinction, when running collection groups on `collections`, we would not only be reading all user's collections, but on public one's as well.

firebase/src/core/data/gateways/firestore-collection-name.ts
matuella: That's because it groups ALL firestore collections (including sub-collections) using its names.

firebase/src/core/data/gateways/firestore-collection-name.ts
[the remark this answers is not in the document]
matuella: Not updated, will try to update it and then let u know.

firebase/src/core/data/gateways/firestore-gateway.ts
matuella: Given my past experiences, I've never instantiated multiple gateways. Encapsulating the transaction lifecycle gives much more freedom and less verbosity to this implementation, while providing the same functionality if using an external "controller" to this lifecycle.

firebase/src/core/data/gateways/firestore-collection-name.ts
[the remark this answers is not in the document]
matuella: Updated with the missing `memosOrder` property, in `Collection Fields`.

firebase/src/core/data/gateways/firestore-collection-name.ts
matuella: I can't remember why I did this, as I've had a hiatus of about 6 weeks on this project, that's completely on me. My bad. I will rollback to something that still make sense.

I think it was related to a possible future marketplace feature, but it's so far-fetched that it makes no sense to do something like that right now

firebase/src/core/data/gateways/firestore-gateway.ts
ggirotto: I'm pointing some potential issues that we already faced when using multiple transactions. Consider the example of running multiple transactions in parallel:

```typescript

await Promise.all([
  firestore.runTransaction(() => {
    // Do some stuff here  
  }),
  firestore.runTransaction(() => {
    // Do other stuff here  
  }),
])

```

The transactions would overwrite each other, since the same gateway is being used. This is the only use-case limitation that I faced when using a transaction inside the gateway.

matuella: I think this is a problem with the implementation usage itself, and not the logic.

Maybe we will eventually face the same problems, but as it stands right now, there is no need to add more complexity unless we know we will need it.

matuella:

firebase/src/core/data/gateways/firestore-gateway.ts
ggirotto: Agreed, just wanted to state because we faced in the past. If we start doing heavy jobs and parallel transactions, we'll need to rethink how to handle with this transaction (and probably use providers to ensure only one transaction is used). But I agree about keeping it simple for now, but we must keep this specific problem in mind.

ggirotto: Agreed, just wanted to state because we faced in the past. If we start doing heavy jobs and parallel transactions, we'll need to rethink how to handle with this transaction (and probably use providers to ensure only one transaction is used). But I agree about keeping it simple for now, but we must keep this specific problem in mind.

ggirotto (approved):

firebase/src/core/data/gateways/firestore-gateway.ts
matuella: I also think that there are better solutions to handle the transaction management by not exposing it to a parent, but that's a completely different discussion

matuella:

PR #204, opened 2021-09-27
Add `FirestoreGateway.getCollection` tests

matuella:


firebase/test/core/data/gateways/firestore-gateways.spec.ts
```diff
@@ -108,5 +115,59 @@ describe("FirestoreGateway", () => {
         assert(transactionMock.get.calledOnce);
       });
     });
+
+    describe("getCollection", () => {
```
ggirotto, line 119: NIT: This suggestion apply to all next test PR's: I think they could be split into separate files. Although they belong to the same "context" opening a 400+ lines file is not pleasant at all, even when things are nested as the test structures are.

ggirotto (approved):

firebase/test/core/data/gateways/firestore-gateways.spec.ts
matuella: Looking at third-party open-sourced projects, that's pretty standard of unit testing, but please correct me if I'm wrong. I think that's up to your usage of the IDE's or even vim/emacs. This file is simply insanely verbose because of the amount of mocking/stubbing.

Slightly offtopic: if we argue about vscode, which is the standard that we use in memo, I've learned to navigate using symbols pretty decently, so I am completely fine with having files that have 300+ lines if they are coherent and have a clear responsibility, given the "architecture".

<details>
<summary>Embedded symbol navigation</summary>

Screen Shot 2021-09-27 at 17 13 07

</details>

[If you press `CMD/CTRL + SHIFT + O`, you'll open the symbols navigation, which you can quickly filter/navigate to specific properties, functions and whatnot.
](https://code.visualstudio.com/Docs/editor/editingevolved#_go-to-symbol)

<details>
<summary>Filtered symbol navigation</summary>

Screen Shot 2021-09-27 at 17 18 44

</details>

matuella:
Asked your review again, as I don't think your point is a NIT, so I had to add some personal opinions on this matter

ggirotto:
> Asked your review again, as I don't think your point is a NIT, so I had to add some personal opinions on this matter

I still think this is a NIT because I personally don't use symbols for navigate - they're definitely better, I am just not used to - and I prefer smaller files with less amount of lines. I can't argue that it's better than having everything in a single file - which also have the point of not duplicating stubs as you mentioned -, it's literally my personal preference about it.

ggirotto (approved):

matuella:
> Ideally using VSCode symbols + awesome testing descriptions would make the files sizes insignificant, I'm just not used to use it, and maybe I should.

Ideally, files should be way more compact and contained, but there are use-cases such as these that requires a lot of verbosity, and I advocate for us to not simply split into smaller files just for the sake for splitting it, because we can still deal with it fine, although they'll never be as good as a small files that have between 50 to 200 lines (quite obvious).

matuella:

Issue #224, opened 2021-10-29
[Component] Custom Buttons


ggirotto:
### Summary

The buttons used in `memo` app doesn't follow the [Material's guidelines](https://m3.material.io/components/buttons/guidelines). This forces us to create custom buttons `Widget`s that reproduce this custom behavior.

### Business Rules

These buttons should respect the visual behavior specified in the Figma project, which includes 3 possible states:

- `normal` - The normal state of the button, without any kind of interaction.
- `pressed` - When the button is pressed.
- `disabled` - When the button doesn't have a `onTap` action.

The buttons must also allow a `trailing` Widget, which should be customizable and must follow the same color as the button text.

It's important to notice that these buttons will probably not use those that Flutter team has built when developing Flutter SDK, which means that we lose features that are inherited when using the default buttons, like accessibility as an example. The developer should study the Flutter default buttons implementation and preserve at most these features.

### Screenshots

<details>
<summary>ElevatedButton</summary>

![CleanShot 2021-10-29 at 09 19 11](https://user-images.githubusercontent.com/11745745/139433095-9aa96315-e719-44f2-bdb6-6ab5dfc48567.png)

</details>

<details>
<summary>Secondary ElevatedButton</summary>

![CleanShot 2021-10-29 at 09 19 33](https://user-images.githubusercontent.com/11745745/139433126-bf3b6b99-6b89-4a28-8d82-649da3d02005.png)

</details>

<details>
<summary>TextButton</summary>

![CleanShot 2021-10-29 at 09 19 55](https://user-images.githubusercontent.com/11745745/139433219-11536de1-22b0-4457-8800-cb32d98b4b4e.png)

</details>

PR #226, opened 2021-10-29
Creates custom buttons

lucasbiancogs:
*Faça aqui uma breve descrição das alterações que esse PR faz e o porquê / Describe the changes proposed in this PR and their motivation:*

Creates visually opinionated buttons that that will be used throughout the application. This is necessary because the buttons used in memo app doesn't follow the Material's guidelines.

*Liste pelo menos um issue relacionado à este PR. / List at least one open issue related to this PR:*

Related to [[Component] Custom Button](https://github.com/olmps/memo/issues/224).

## Pre-launch Checklist

- [x] Eu li o [CONTRIBUTING.md](../CONTRIBUTING.md) e segui as sugestões descritas ali sobre boas práticas de código / I
read [CONTRIBUTING.md](../CONTRIBUTING.md) and followed the suggestions proposed there about writing good code;
- [x] Eu listei pelo menos um issue relacionado à este PR / I listed at least one issue related to this PR;
- [x] Eu adicionei novos testes para cobrir as mudanças que eu fiz neste PR, ou este PR é uma exceção / I added new 
tests that cover the changes made in this PR or this PR is a test-exception;
- [x] Eu atualizei o [CHANGELOG.md](./../CHANGELOG.md) com as mudanças feitas neste PR / I updated 
[CHANGELOG.md](./../CHANGELOG.md) with the changes made in this PR;
- [x] Eu atualizei/adicionei comentários relevantes nas modificações efetuadas / I updated or added relevant docs in the
modifications I've made;
- [x] Todos os testes previamente existentes e os novos criados por mim estão finalizando com sucesso / All the existing
tests - the old and the new ones - are finishing with success.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,14 +10,20 @@ a beta or production release, they must be documented here).
 
 ## [Unreleased]
 
+### Added
+
+- Visually Opinionated Buttons (Primary, Secondary and Text).
+
 ## [0.2.0] - 2021-08-13
 
 ### Updated
+
 - Several `ExecutionTerminal` enhancements:
   - Allows the user to go back/forward on the same memo.
   - The selected difficulty don't require a confirmation anymore - once it's selected, it animates to the next memo.
 
 ### Fixed
+
 - Several `ExecutionTerminal` fixes:
   - Actions overlapping in smaller devices.
   - Inconsistent state when memos were executed too fast.
@@ -25,36 +31,40 @@ a beta or production release, they must be documented here).
 - Replaced oval gradient in `CollectionCard` with `LinearGradient`, attempting to solve performance issues.
 
 ### Removed
+
 - Hero animation from `CollectionCard`.
 
 ## [0.2.0-rc.1] - 2021-08-02
 
 ### Added
+
 - Firebase Analytics SDK with no need for consent - disabled `AdId` collection.
 - A proper empty state for `CollectionsPage` - the `_CollectionsEmptyState`.
 - `DestructiveButton`, a customized button for destructive operations - with custom layout specs.
 - `MemoThemeData.destructiveSwatch` and a its respective destructive swatch for the classic theme.
 
 ### Updated
+
 - Android & iOS Fastfiles to automatically distribute uploaded builds to external testers.
 - Both `pull-request` and `release` workflows now:
   - Caches flutter's SDK.
   - Run Flutter tests only once.
   - Strategy used to authenticate with `GoogleService-Info` (only required in `release`, using an empty template in
-  `pull-request`).
+    `pull-request`).
   - Uses the Flutter's beta channel + latest version, requirements of the animation jank fix.
 - Android `minSdkVersion` from `16` to `17` and `multiDexEnabled` to `true`, requirements of `flutter-quill` dependency.
 - iOS `MinimumOSVersion` from `8.0` to `9.0`, an apparent dependency on the latest beta channel version.
 - Flutter's SDK to `2.4.0-4.2.pre` to fix the animation jank.
 - `Fastfile` to use `--bundle-sksl-path` option, requirements of the animation jank fix.
 - When quitting the execution, `CollectionExecutionPage` now uses the `showSnappableDraggableModalBottomSheet` instead
-of `AlertDialog`.
+  of `AlertDialog`.
 
 #### Collections
 
 - Cohesion improvements on `comecando_com_git`.
 
 ### Fixed
+
 - Discord link was expired, now it's a permalink.
 - White screen before `SplashPage` was loaded (using `flutter_native_splash` and generating native splash screens).
 - `AssetIconButton` wasn't conforming to the correct layout specs.
@@ -66,23 +76,26 @@ of `AlertDialog`.
 ## [0.1.0] - 2021-07-16
 
 ### Added
+
 - `bdd_fundamentos_01`, `fundamentos_scrum`, `guia_scrum`, `kotlin_fundamentos_01` and `manifesto_agil` collections.
-Thanks to all contributors!
+  Thanks to all contributors!
 
 ## [0.1.0-rc.3] - 2021-07-15
 
 ### Added
+
 - `Contributor` model and its `ContributorSerializer` serializer.
 - `EnvMetadata` and its respective implementation that provides application's environment constants.
 - `SettingsSection.community` with `LinkSettingsItem` to discord.
 - `showSnappableDraggableModalBottomSheet` utility.
 - `MultiContributorsView` and `SingleContributorView` widgets.
 - `Firebase` and `FirebaseCrashlytics` to record unexpected crashes, errors and exceptions.
-  
+
 ### Updated
+
 - All files documentations, including standardizing communication.
 - `CollectionMemos` and `Collection` now have a `contributors` property, exposing all associated contributors with that
-particular collection.
+  particular collection.
 - `LoadedCollectionDetailsState` now also provides a list of `ContributorInfo` associated with that `Collection`.
 - `LinkButton` widget now exposes `backgroundColor` and `textStyle` properties.
 - `ExternalLinkButton` widget now exposes `iconColor`, `backgroundColor` and `textStyle` properties.
@@ -91,35 +104,40 @@ particular collection.
 - Existing collections with their respective contributors.
 
 ### Fixed
+
 - Missing `SafeArea` in `Scaffold.bottomNavigationBar` for devices with home indicator.
 - Missing `SettingsVM` interface.
-- Hero animations built through  `buildHeroCollectionCardFromItem` weren't using an unique `Hero.tag`.
+- Hero animations built through `buildHeroCollectionCardFromItem` weren't using an unique `Hero.tag`.
 
 ## [0.1.0-rc.2] - 2021-05-12
 
 ### Added
+
 - `comecando_com_git`, `ecossistema_do_flutter` and `swift_fundamentos_01` collections.
 
 ### Removed
+
 - `git_getting_started` collection.
 
 ### Updated
+
 - Added new resources to `resources.json`.
 
 ### Fixed
+
 - Fixed [`release`](.github/workflows/release.yml) workflow to use personal access token when pushing changes to the
-repo.
+  repo.
 
 ## [0.1.0-rc.1] - 2021-05-08
 
 Introduces the first release candidate with all the first idealized functionalities for `memo`:
 
 - Collections listing, which sections all collections using its category and segments the contents by explore and
-review:
+  review:
   - Explore shows all collections that haven't all memos completed at least once;
   - Review shows all collections that have all memos completed at least once.
 - Progress, showing the user's metadata for application-wide collection's executions, such as time spent, the
-percentage for each answer, etcetera;
+  percentage for each answer, etcetera;
 - Collection details and execution, which allows the user to study upon a compiled collection in this repository;
 - Settings page with info related to this project.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,19 +11,16 @@ a beta or production release, they must be documented here).
 ## [Unreleased]
 
 ### Added
-
 - Visually Opinionated Buttons (Primary, Secondary and Text).
 
 ## [0.2.0] - 2021-08-13
 
 ### Updated
-
 - Several `ExecutionTerminal` enhancements:
   - Allows the user to go back/forward on the same memo.
   - The selected difficulty don't require a confirmation anymore - once it's selected, it animates to the next memo.
 
 ### Fixed
-
 - Several `ExecutionTerminal` fixes:
   - Actions overlapping in smaller devices.
   - Inconsistent state when memos were executed too fast.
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -10,51 +10,61 @@ a beta or production release, they must be documented here).
```
ggirotto, line 10: This file spacings and new lines were automatically added?

lib/application/widgets/theme/custom_button.dart
```diff
@@ -0,0 +1,309 @@
...
+      leadingWidgetBuilder: leadingAsset != null ? leadingAssetBuilder : null,
+      onPressed: onPressed,
+      textStyleBuilder: (state) => textTheme.copyWith(color: buttonColor(state)),
+    );
+  }
+}
+
+/// Wraps a [CustomButton] to become similar to a [ElevatedButton].
+///
+/// Requires a [backgroundColorBuilder] to build accordingly to the button state.
```
ggirotto, line 142:
```suggestion
/// Wraps a [CustomButton] following a similar style to [ElevatedButton].
///
/// Reused by custom buttons implementations. See also:
///   * [PrimaryElevatedButton]
///   * [SecondaryElevatedButton]
```

lib/application/widgets/theme/custom_button.dart
```diff
@@ -0,0 +1,309 @@
...
+}
+
+/// Shared visual style between all custom button's.
+///
+/// Provides a common and customizable button style that must be used by all custom buttons.
+///
+/// See also:
+///
+///   * [PrimaryElevatedButton] - The primary visually opinionated alternative to [ElevatedButton].
+///   * [SecondaryElevatedButton] - The visually opinionated alternative to [ElevatedButton].
```
ggirotto, line 194:
```suggestion
///   * [SecondaryElevatedButton] - The secondary visually opinionated alternative to [ElevatedButton].
```

ggirotto (changes requested):

ggirotto:
@matuella Ready to review

ggirotto (dismissed):

lib/application/widgets/theme/custom_button.dart
```diff
@@ -0,0 +1,313 @@
...
+
+    final buttonContent = InkWell(
+      // Disables default splash/highlight animation
+      splashColor: Colors.transparent,
+      highlightColor: Colors.transparent,
+      onTap: widget.onPressed,
+      onHighlightChanged: (isPressed) {
+        setState(() {
+          state = isPressed ? ButtonState.pressed : ButtonState.normal;
+        });
```
matuella, line 293: We should use hooks here

lib/application/widgets/theme/custom_button.dart
```diff
@@ -0,0 +1,313 @@
...
+    state = widget.onPressed != null ? ButtonState.normal : ButtonState.disabled;
+    super.initState();
+  }
+
+  @override
+  void didUpdateWidget(covariant CustomButton oldWidget) {
+    if (widget.onPressed != oldWidget.onPressed) {
+      setState(() {
+        state = widget.onPressed != null ? ButtonState.normal : ButtonState.disabled;
+      });
```
matuella, line 267: use hooks

lib/application/widgets/theme/custom_button.dart
```diff
@@ -0,0 +1,313 @@
...
+  /// If `null` the default text theme is used.
+  final TextStyle? Function(ButtonState state)? textStyleBuilder;
+
+  @override
+  CustomButtonState createState() => CustomButtonState();
+}
+
+@visibleForTesting
+class CustomButtonState extends State<CustomButton> {
+  late ButtonState state;
```
matuella, line 254: no need to create a stateful widget, you can use a stateless widget with a state hook

lib/application/widgets/theme/custom_button.dart
```diff
@@ -0,0 +1,313 @@
...
+///
+/// Provides a common and customizable button style that must be used by all custom buttons.
+///
+/// See also:
+///
+///   * [PrimaryElevatedButton] - The primary visually opinionated alternative to [ElevatedButton].
+///   * [SecondaryElevatedButton] - The secondary visually opinionated alternative to [ElevatedButton].
+///   * [CustomTextButton] - The visually opinionated alternative to [TextButton].
+@visibleForTesting
+class CustomButton extends StatefulWidget {
```
matuella, line 201: I think keeping this private and testing all implementations (primary, secondary and text) directly would be better here, as the usages of this "internal" behavior is what matters for us. You could test both as well, but I don't know if it would make sense

lib/application/widgets/theme/custom_button.dart
```diff
@@ -0,0 +1,313 @@
...
+
+  /// {@template _CustomElevatedButton.leadingAsset}
+  /// Optional leading asset.
+  ///
+  /// Its color will be overriden by the button `TextStyle`.
+  /// {@endtemplate}
+  final String? leadingAsset;
+
+  /// Builds the button background color based on the button state.
+  final Color? Function(ButtonState state) backgroundColorBuilder;
```
matuella, line 172: When will color be null?

test/utils/widget_pump.dart
```diff
@@ -9,3 +11,15 @@ Future<void> pumpMaterialScoped(WidgetTester tester, Widget widget) => tester.pu
...
+Future<void> pumpMaterialScopedWithTheme(WidgetTester tester, Widget widget) => tester.pumpWidget(
+      ProviderScope(
+        overrides: [themeController.overrideWithValue(ThemeController())],
+        child: MaterialApp(
+          home: Scaffold(
+            body: widget,
+          ),
+        ),
+      ),
+    );
```
matuella, line 25:
```suggestion
/// Wraps a [pumpMaterialScoped] with a [ProviderScope].
Future<void> pumpProviderScoped(WidgetTester tester, Widget widget, [ List<?> overrides ]) => tester.pumpWidget(
      ProviderScope(
        overrides: overriddes ?? [],
        child: MaterialApp(
          home: Scaffold(
            body: widget,
          ),
        ),
      ),
    );
```

I think there is no need to override the `ThemeController`, as the `themeController` already pass the exact same implemented value that you provided on the override.

Also I think allowing an arbitrary list of overrides would be best (don't know the exact type, so I've put a `?`).

matuella (commented):
It's important to track the removal of all previous button-related material's theme customizations and update the existing buttons.

lib/application/widgets/theme/custom_button.dart
ggirotto: We are using `StatefulWidget` because we need to update the button's state based on `void didUpdateWidget(covariant CustomButton oldWidget)` event. I'm not sure if we can achieve such behavior with hooks.

lib/application/widgets/theme/custom_button.dart
ggirotto: To do that we would have to change the testing strategy to detect the state change based on the Widget interface update, because making `CustomButton` makes us loose access to its `state` property. The problem that I see here is that if we decide to change the buttons interface it would directly affect the tests, while testing by `CustomButton.state` test the behavior itself, independent from the UI

test/utils/widget_pump.dart
ggirotto: I'm not sure what you suggested about removing the `ThemeController` override, but without it it's not possible to test the widget because it uses the hooks to fetch the controller from the Widgets tree. If I misunderstood the suggestion please let me know.

matuella: > it uses the hooks to fetch the controller from the Widgets tree

It's provided by [default by `themeController`](https://github.com/olmps/memo/blob/custom-button/lib/application/theme/theme_controller.dart#L10) (variable containing the theme provider).

lib/application/widgets/theme/custom_button.dart
matuella: This is a `_CustomElevatedButton`, how come `CustomTextButton` has anything to do with it?

lib/application/widgets/theme/custom_button.dart
matuella: Hooks can achieve anything a stateful widget can, just look at any example in this repository or the `flutter_hooks` itself.

test/utils/widget_pump.dart
matuella: > it uses the hooks to fetch the controller from the Widgets tree

It's provided by [default by `themeController`](https://github.com/olmps/memo/blob/custom-button/lib/application/theme/theme_controller.dart#L10) (variable containing the theme provider).

lib/application/widgets/theme/custom_button.dart
matuella: And that's fine, because unit tests are affordable to be deleted/updated/added, but we want to make sure that the functionality itself is working as intended, and not some particular reusable internal state, but I do see where you're getting at, and, if necessary, I think it could be solved by structuring the tests where some of its logic could be shared/reused by all of these interfaces

test/utils/widget_pump.dart
ggirotto: But we are not using it directly, we use it through [`useTheme`](https://github.com/olmps/memo/blob/302ead3a120f1b10a952397ebbcec4f089dd3d2e/lib/application/theme/theme_controller.dart#L90) which trie to read it from the context scope.

lib/application/widgets/theme/custom_button.dart
```diff
@@ -0,0 +1,276 @@
...
+  @override
+  Widget build(BuildContext context) {
+    final initialState = onPressed != null ? _ButtonState.normal : _ButtonState.disabled;
+    final state = useState(initialState);
+
+    useEffect(() {
+      if (state.value != _ButtonState.pressed) {
+        state.value = onPressed != null ? _ButtonState.normal : _ButtonState.disabled;
+      }
+    });
```
matuella, line 241: Where this particular scenario occurs?

matuella: I think this is a bad design of some sort, not sure though, but it's odd to me that an effect must run every time the build runs

test/utils/widget_pump.dart
```diff
@@ -9,3 +11,16 @@ Future<void> pumpMaterialScoped(WidgetTester tester, Widget widget) => tester.pu
...
+    tester.pumpWidget(
+      ProviderScope(
+        overrides: List.from(overrides)..add(themeController.overrideWithValue(ThemeController())),
+        child: MaterialApp(
+          home: Scaffold(
+            body: widget,
+          ),
+        ),
+      ),
+    );
```
matuella, line 26:
```suggestion
/// Wraps a [pumpMaterialScoped] with a [ProviderScope] that optionally uses [overrides].
Future<void> pumpThemedProviderScoped(WidgetTester tester, Widget widget, [List<Override> overrides = const []]) =>
    tester.pumpWidget(
      ProviderScope(
        overrides: overrides,
        child: MaterialApp(
          home: Scaffold(
            body: widget,
          ),
        ),
      ),
    );
```

Make the callers pass what they want.

Also:

> But we are not using it directly, we use it through useTheme which trie to read it from the context scope.

It's that once you give a `ProviderScope`, it knows how to access the `themeController`, because just like in the application, we don't override anything from the theme, we use the provider-provided one.

Have you tried running these tests WITHOUT overriding the `ThemeController`, which is provided by default https://github.com/olmps/memo/blob/9a9a8197fd0f8fe3af0213a408bcc166462aea81/lib/application/theme/theme_controller.dart#L10 ?

test/application/widgets/theme/custom_button_test.dart
```diff
@@ -0,0 +1,205 @@
...
+import '../../../utils/widget_pump.dart';
+
+void main() {
+  group('PrimaryElevatedButton -', () {
+    const buttonColor = Colors.red;
+    final normalColor = buttonColor.shade400;
+    const pressedColor = buttonColor;
+    final disabledColor = buttonColor.withOpacity(0.4);
+
+    testWidgets('should inits with normal background color when onPressed is not null', (tester) async {
```
matuella, line 15: what is inits?

test/application/widgets/theme/custom_button_test.dart
```diff
@@ -0,0 +1,205 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:memo/application/constants/images.dart' as images;
+import 'package:memo/application/widgets/theme/custom_button.dart';
+
+import '../../../utils/widget_pump.dart';
+
+void main() {
```
matuella, line 8: Most of these tests are copy-pasted. Can't we simply test these buttons within a common logic?

lib/application/widgets/theme/custom_button.dart
ggirotto: That's the only way I found of replacing the old `void didUpdateWidget(covariant CustomButton oldWidget)` functions, since it's now a `HookWidget`. I haven't found further discussions about how to perform this replacement in the Riverpod repository as well, so I'm not sure if there's something obvious that I'm not seeing or if in this scenario we should be using `StatefulWidget` instead. We need to update the buttons interface once the user taps it, and the only hook that I though it would be able to do that is the `useEffect`, since the tap triggers a rebuild.

test/utils/widget_pump.dart
ggirotto: Now I get it. It works without the `ThemeController` override, just with the normal `ProviderScope.override`. Now sure what I've tested before that didn't work (I think I was removing the entire `ProviderScope`).

test/application/widgets/theme/custom_button_test.dart
ggirotto: `init`

test/application/widgets/theme/custom_button_test.dart
ggirotto: Added auxiliary functions, let me know what do you think

test/application/widgets/theme/custom_button_test.dart
ggirotto: `init`

lib/application/widgets/theme/custom_button.dart
```diff
@@ -0,0 +1,276 @@
...
+      enabled: state.value != _ButtonState.disabled,
+      child: Container(
+        height: dimens.minButtonHeight,
+        decoration: BoxDecoration(
+          color: backgroundColorBuilder?.call(state.value),
+          borderRadius: dimens.genericRoundedElementBorderRadius,
+        ),
+        child: buttonContent,
+      ),
+    );
```
matuella, line 274:
```suggestion
  Widget build(BuildContext context) {
    final state = useState(stateForPress(false));

    _ButtonState stateForPress(bool isPressed) {
      if (onPressed == null) {
        return _ButtonState.disabled;
      }
      
      return isPressed ? _ButtonState.pressed: _ButtonState.normal;
    };

    final rowWidgets = <Widget>[];

    if (leadingWidgetBuilder != null) {
      rowWidgets.addAll([
        leadingWidgetBuilder!(state.value),
        context.horizontalBox(Spacing.xSmall),
      ]);
    }

    rowWidgets.add(Text(text, style: textStyleBuilder?.call(state.value)));

    final buttonContent = InkWell(
      // Disables default splash/highlight animation
      splashColor: Colors.transparent,
      highlightColor: Colors.transparent,
      onTap: onPressed,
      onHighlightChanged: (isPressed) => state.value = stateForPress(isPressed),
      child: Center(child: Row(mainAxisAlignment: MainAxisAlignment.center, children: rowWidgets)),
    );

    return Semantics(
      button: true,
      enabled: state.value != _ButtonState.disabled,
      child: Container(
        height: dimens.minButtonHeight,
        decoration: BoxDecoration(
          color: backgroundColorBuilder?.call(state.value),
          borderRadius: dimens.genericRoundedElementBorderRadius,
        ),
        child: buttonContent,
      ),
    );
```

matuella (approved):

Issue #214, opened 2021-10-27
[Collection Creation] - Adds collection creation


ggirotto:
Central issue to track the [Collection creation](https://github.com/olmps/memo/projects/8) for `memo`.

### Why

User must be able to create and edit its own collections in the app, otherwise it would be constrained by the collections that are created by the community.

This functionality gives the opportunity for users to create and maintain specific collections, given how they see fit.

### How

The collections creation should be considered completed after all issues below are closed.

- [x] https://github.com/olmps/memo/issues/215
- [x] https://github.com/olmps/memo/issues/216
- [x] https://github.com/olmps/memo/issues/217
- [x] https://github.com/olmps/memo/issues/218
- [x] https://github.com/olmps/memo/issues/219
- [x] https://github.com/olmps/memo/issues/220
- [x] https://github.com/olmps/memo/issues/221
- [x] https://github.com/olmps/memo/issues/222

This will be worked on [`update-collection` branch](https://github.com/olmps/memo/tree/update-collection), which will be eventually be merged to `main` when we consider it stable.

PR #227, opened 2021-11-03
Updates `pubspec.yaml` and workflows versions to Flutter 2.5.0

ggirotto:


pubspec.yaml
```diff
@@ -48,7 +48,7 @@ dependencies:
   hooks_riverpod: ^0.13.0
 
   # Keep dependency locked, as we need it to be the exact same in `memo-editor`
-  flutter_quill: 1.3.3
+  flutter_quill: 2.0.14
```
matuella, line 51: I prefer to not bump this.

matuella (commented):
Side-effects of bumping `flutter_quill` were evaluated?

ggirotto:
> Side-effects of bumping `flutter_quill` were evaluated?

Yes. It's working as expected in the collections executions workflow.
We may have some news while developing the [collections creation](https://github.com/olmps/memo/issues/214) feature, but I'm confident since we only need the basics for this workflow (bold, italic, underline and code)

matuella (approved):

PR #229, opened 2021-11-04
`RichTextField` component

ggirotto:


lib/application/widgets/theme/rich_text_field.dart
```diff
@@ -0,0 +1,297 @@
...
+    final inputDecorationTheme = Theme.of(context).inputDecorationTheme;
+
+    final initialText = '${controller?.richTextContent ?? ''}\n';
+    final initialController = quill.QuillController(
+      document: quill.Document.fromDelta(quill.Delta()..insert(initialText)),
+      selection: const TextSelection.collapsed(offset: 0),
+    );
+    // Holds the entire field `quill` controller which is later replaced by `modalEditorController`, since it's not
+    // possible to forward the content from one controller to another.
+    final richFieldController = useState(initialController);
```
ggirotto, line 48: I'm not sure about this, it seems completely bizarre, although I couldn't find any other way to forward `modalEditorController` content to `richFieldController`

lib/application/widgets/theme/rich_text_field.dart
```diff
@@ -0,0 +1,297 @@
...
+        if (keyboardHeight > 0) ...[
+          _RichTextFieldToolbar(controller),
+          SizedBox(height: keyboardHeight),
+        ]
+      ],
+    );
+  }
+}
+
+/// Provides a custom styled and reusable [quill.QuillEditor].
```
matuella, line 175: is this doc linking to the right class?

lib/application/widgets/theme/rich_text_field.dart
```diff
@@ -0,0 +1,297 @@
...
+  final Color backgroundColor;
+  final String placeholder;
+  final bool showCursor;
+  final bool readonly;
+
+  @override
+  Widget build(BuildContext context) {
+    final theme = useTheme();
+    final textTheme = Theme.of(context).textTheme;
+    const zeroTuple = Tuple2<double, double>(0, 0);
```
matuella, line 200: care to explain this?

lib/application/widgets/theme/rich_text_field.dart
```diff
@@ -0,0 +1,297 @@
...
+    _ToolbarAttribute.italic: images.italic,
+    _ToolbarAttribute.underline: images.underline,
+    _ToolbarAttribute.code: images.code,
+  };
+
+  @override
+  Widget build(BuildContext context) {
+    final theme = useTheme();
+
+    final selectedAttributes = useState<List<_ToolbarAttribute>>([]);
```
matuella, line 262: Ideally should be a Set. If order is required, you should guarantee unique items in the list or provide a way to sort the Set

matuella (commented):
This `false` shrink behavior to some buttons but `true` for others is really confusing. I think it's expected for all buttons to work the same (meaning all shrink, which is flutter's default if I'm not mistaken) and, if there are specific stretch scenarios, use shrink to false. 

---

Also, I think that the "editor" behavior should be all comprised within a provider and separated from the UI - the widgets are really bloated and are assuming much more than simply building the UI, such as:

```
    final content = richFieldController.value.plainTextEditingValue.text;
    // `quill` automatically adds `\n` when initializing the editor without content
    final hasContent = content.isNotEmpty && content != '\n';
```

---

And at last, it seems to me that this editor has a lot of use-cases to be tested that are completely ignored, such as a ton of use-cases using the attributes.

ggirotto:
> This false shrink behavior to some buttons but true for others is really confusing. I think it's expected for all buttons to work the same (meaning all shrink, which is flutter's default if I'm not mistaken) and, if there are specific stretch scenarios, use shrink to false.

I tested the scenarios using `ElevatedButton` and `TextButton` and the TLDR is that the elevated expands to the maximum available space, while the text uses the minimum space as necessary:

<details>
<summary>Screenshot</summary>

CleanShot 2021-11-09 at 11 13 16@2x

</details>

Given these tests it makes sense to use this internal `shrink` behavior, since all buttons share from the same `_CustomButton` structure.

lib/application/widgets/theme/custom_button.dart
```diff
@@ -292,7 +298,11 @@ class CustomButtonState extends State<CustomButton> {
           state = isPressed ? ButtonState.pressed : ButtonState.normal;
         });
       },
-      child: Center(child: Row(mainAxisAlignment: MainAxisAlignment.center, children: rowWidgets)),
+      child: Row(
+        mainAxisAlignment: MainAxisAlignment.center,
+        mainAxisSize: widget.shrink ? MainAxisSize.min : MainAxisSize.max,
+        children: [context.horizontalBox(Spacing.small), ...rowWidgets, context.horizontalBox(Spacing.small)],
```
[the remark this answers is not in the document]
ggirotto: Actually based on the tests made here https://github.com/olmps/memo/pull/229#issuecomment-964196783 I would even remove this paddings, since they should be defined by the buttons parent if necessary (since we've scenarios where we wouldn't like to have any implicit padding)

lib/application/widgets/theme/rich_text_field.dart
```diff
@@ -0,0 +1,297 @@
...
+import 'package:memo/application/widgets/theme/custom_button.dart';
+import 'package:memo/application/widgets/theme/themed_container.dart';
+import 'package:tuple/tuple.dart';
+
+/// A [TextField] variation that supports rich text content.
+///
+/// The field supports rich text content by opening a modal when tapped with an embedded `Quill Editor`.
+///
+/// The field also supports presenting rich text content, different from [TextField] which only supports simple
+/// text content. The field content height is limited by [dimens.richTextFieldConstraints].
```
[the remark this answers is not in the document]
ggirotto: What about

```dart
/// A [TextField] that follows most of WYSIWYG editors functionality.
///
/// Supports presenting rich text content and opens a modal with a rich text editor when tapped.
```

lib/application/widgets/theme/custom_button.dart
```diff
@@ -257,7 +263,15 @@ class _CustomButton extends HookWidget {
       highlightColor: Colors.transparent,
       onTap: onPressed,
       onHighlightChanged: (isPressed) => state.value = isPressed ? _ButtonState.pressed : _ButtonState.normal,
-      child: Center(child: Row(mainAxisAlignment: MainAxisAlignment.center, children: rowWidgets)),
+      child: Row(
```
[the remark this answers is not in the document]
ggirotto: https://github.com/olmps/memo/pull/229#discussion_r745663521

lib/application/widgets/theme/rich_text_field.dart
```diff
@@ -0,0 +1,297 @@
...
+  Widget build(BuildContext context) {
+    final theme = useTheme();
+    final textTheme = Theme.of(context).textTheme;
+    final inputDecorationTheme = Theme.of(context).inputDecorationTheme;
+
+    final initialText = '${controller?.richTextContent ?? ''}\n';
+    final initialController = quill.QuillController(
+      document: quill.Document.fromDelta(quill.Delta()..insert(initialText)),
+      selection: const TextSelection.collapsed(offset: 0),
+    );
```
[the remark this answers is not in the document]
ggirotto: Created a custom hook to use with `QuillController`

lib/application/widgets/theme/rich_text_field.dart
ggirotto: I figured out that having two controllers didn't make any sense at all. I just need to use the same `QuillController` in both instancies.

lib/application/widgets/theme/rich_text_field.dart
ggirotto: Yes

lib/application/widgets/theme/rich_text_field.dart
ggirotto: Added docs

lib/application/widgets/theme/rich_text_field.dart
```diff
@@ -0,0 +1,297 @@
...
+  /// Controls the text content being edited.
+  final RichTextEditingController? controller;
+
+  @override
+  Widget build(BuildContext context) {
+    final theme = useTheme();
+    final textTheme = Theme.of(context).textTheme;
+    final inputDecorationTheme = Theme.of(context).inputDecorationTheme;
+
+    final initialText = '${controller?.richTextContent ?? ''}\n';
```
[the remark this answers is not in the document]
ggirotto: Removed (not necessary)

lib/application/widgets/theme/rich_text_field.dart
```diff
@@ -0,0 +1,343 @@
...
+  const RichTextField({required this.modalTitle, required this.placeholder, this.controller, this.focus});
+
+  /// The modal title positioned in the top-left corner of the modal editor.
+  final Widget modalTitle;
+
+  /// The placeholder used in the rich text editor when it has no content.
+  final String placeholder;
+
+  /// Controls the text content being edited.
+  final TextEditingController? controller;
```
ggirotto, line 33: What do you think about using the known `TextEditingController` instead a custom one? I realized we have the same use-cases that a normal textfield has. The difference is that our `text` property will be the rich text JSON instead a normal text.

lib/application/widgets/theme/rich_text_field.dart
```diff
@@ -0,0 +1,343 @@
...
+
+  @override
+  quill.QuillController build(BuildContext context) => _controller;
+
+  @override
+  void dispose() => _controller.dispose();
+
+  @override
+  String get debugLabel => 'useQuillController';
+}
```
ggirotto, line 343: Created a custom hook to control `quill.QuillController` lifecycle in the StatelessWidget. What do you think?

lib/application/widgets/theme/custom_text_field.dart
```diff
@@ -114,7 +118,7 @@ class CustomTextField extends HookConsumerWidget {
 
     final textField = Container(
       decoration: BoxDecoration(
-        border: _hasErrorText ? Border.all(color: theme.destructiveSwatch, width: dimens.genericBorderHeight) : null,
+        color: _hasErrorText ? theme.destructiveSwatch : null,
```
ggirotto, line 121: Had some animation issues when using `border`. Using just `color` with `borderRadius` solved the problem (not sure why, but I gave up while trying to understand)

lib/application/hooks/rich_text_field_controller_hook.dart
```diff
@@ -0,0 +1,45 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_hooks/flutter_hooks.dart';
+import 'package:memo/application/widgets/theme/rich_text_field.dart';
+
+const useRichTextEditingController = _RichTextEditingControllerHookCreator();
```
matuella, line 5: missing docs

lib/application/widgets/theme/custom_text_field.dart
matuella: If they achieve the same layout that's fine, but does it though?

test/application/widgets/theme/custom_text_field_test.dart
```diff
@@ -116,12 +116,11 @@ void main() {
...
-        final containerBorder = containerDecoration.border! as Border;
 
-        expect(containerBorder.top.color, expectedColor);
-        expect(containerBorder.right.color, expectedColor);
-        expect(containerBorder.bottom.color, expectedColor);
-        expect(containerBorder.left.color, expectedColor);
+        expect(containerDecoration.color, expectedColor);
+        expect(containerDecoration.color, expectedColor);
+        expect(containerDecoration.color, expectedColor);
+        expect(containerDecoration.color, expectedColor);
```
matuella, line 123: what

test/utils/widget_pump.dart
```diff
@@ -16,10 +17,11 @@ Future<void> pumpProviderScoped(WidgetTester tester, Widget widget, [List<Overri
...
       ProviderScope(
         overrides: overrides,
-        child: MaterialApp(
-          home: Scaffold(
-            body: widget,
-          ),
-        ),
+        child: Builder(builder: (context) {
+          return MaterialApp(
+            theme: ThemeController().currentThemeData(context),
```
matuella, line 22: you should probably use the `themeController` provider

lib/application/hooks/rich_text_field_controller_hook.dart
matuella: > Created a custom hook to control quill.QuillController lifecycle in the StatelessWidget. What do you think?

What's the essential difference between this and a `useTextEditingController`?

lib/application/widgets/theme/rich_text_field.dart
```diff
@@ -0,0 +1,436 @@
...
+        );
+
+  String get richText => value.richText;
+  String get plainText => value.plainText;
+  TextSelection get selection => value.selection;
+
+  set richText(String newValue) => value = value.copyWith(richText: newValue);
+  set plainText(String newValue) => value = value.copyWith(plainText: newValue);
+  set selection(TextSelection newValue) => value = value.copyWith(selection: newValue);
+}
```
matuella, line 59: Again, I may be out of the loop here, but what does all this "RichTextFieldController" does that a common "TextFieldController" doesn't? Can't we modify the original one instead of creating a whole new one, with custom hooks and such?

lib/application/widgets/theme/rich_text_field.dart
```diff
@@ -0,0 +1,436 @@
...
+
+    final hasContent = useState(controller?.richText.isNotEmpty ?? false);
+
+    useEffect(() {
+      void editorChanged() {
+        final plainText = quillController.plainTextEditingValue.text.trim();
+
+        controller?.richText = jsonEncode(quillController.document.toDelta().toJson());
+        controller?.plainText = plainText;
+        controller?.selection = quillController.selection;
```
matuella, line 108: This will trigger 3 different notifications, but are the same "update"

lib/application/widgets/theme/rich_text_field.dart
```diff
@@ -0,0 +1,436 @@
...
+        context,
+        ref,
+        child: _RichTextFieldModal(
+          title: modalTitle,
+          controller: quillController,
+          placeholder: placeholder,
+          focus: focus,
+        ),
+      );
+    }
```
matuella, line 128: would move this to a private method of this class

lib/application/widgets/theme/rich_text_field.dart
```diff
@@ -0,0 +1,436 @@
...
+                  showCursor: false,
+                  readonly: true,
+                ),
+              ),
+            ),
+          ] else
+            Text(placeholder, style: textTheme.subtitle1)
+        ],
+      ).withSymmetricalPadding(context, vertical: Spacing.small, horizontal: Spacing.medium),
+    );
```
matuella, line 163: This looks to me like its own widget. Wdyt?

lib/application/widgets/theme/rich_text_field.dart
```diff
@@ -0,0 +1,436 @@
...
+              context.verticalBox(Spacing.xxSmall),
+              Expanded(child: editor),
+              context.verticalBox(Spacing.xxSmall),
+            ],
+          ).withSymmetricalPadding(context, horizontal: Spacing.medium),
+        ),
+        // Only show toolbar action items when the field has focus or when the user has an active selection.
+        if (isFocused.value || hasSelection.value) ...[
+          _RichTextFieldToolbar(controller),
+          SizedBox(height: MediaQuery.of(context).viewInsets.bottom),
```
matuella, line 254: Possibly explain that this is a way to add a bottom space that takes the same height of the currently visible keyboard

lib/application/widgets/theme/rich_text_field.dart
```diff
@@ -0,0 +1,436 @@
...
+    this.showCursor = true,
+    this.readonly = false,
+  });
+
+  final quill.QuillController controller;
+  final Color backgroundColor;
+  final String placeholder;
+  final FocusNode? focus;
+  final bool showCursor;
+  final bool readonly;
```
matuella, line 282: can be explained

lib/application/widgets/theme/rich_text_field.dart
```diff
@@ -0,0 +1,436 @@
...
+}
+
+/// A toolbar of actions to customize the editor content.
+class _RichTextFieldToolbar extends HookConsumerWidget {
+  _RichTextFieldToolbar(this.controller);
+
+  final quill.QuillController controller;
+
+  /// Maps [quill.Attribute] to its respective asset.
+  final Map<quill.Attribute<dynamic>, String> _toolBarAsset = {
```
matuella, line 332: maybe a `Map. unmodifiable`?

lib/application/widgets/theme/rich_text_field.dart
```diff
@@ -0,0 +1,436 @@
...
+  }
+
+  @override
+  quill.QuillController build(BuildContext context) => _controller;
+
+  @override
+  void dispose() => _controller.dispose();
+
+  @override
+  String get debugLabel => 'useQuillController';
```
matuella, line 435:
```suggestion
  String get debugLabel => '_useQuillController';
```

test/application/widgets/theme/custom_text_field_test.dart
ggirotto: My bad

lib/application/widgets/theme/rich_text_field.dart
ggirotto: Updated. Syntax is a little bit worse but it's worth it due to the immutability from the map

lib/application/widgets/theme/rich_text_field.dart
ggirotto: Updated docs and removed `showCursor` in favor of using `readonly`

lib/application/widgets/theme/rich_text_field.dart
ggirotto: I think yes

test/utils/widget_pump.dart
ggirotto: By using `Consumer`, right?

lib/application/widgets/theme/rich_text_field.dart
ggirotto: Actually I started using a common `TextEditingController`. The problems that I found during its usage are:

1. We would have to use `text` to store the string version of the rich content, which is a little bit counter intuitive, since the `text` normally stores plain text when using `TextEditingController`.
2. The `TextEditingController` only supports storing two properties: `text` (String) and `selection` (TextSelection), but we also need to have access to the plain text version of the rich text editor. Because of that necessity and further ones that may happen, I think that having a custom controller gives us flexibility to add/remove necessary information.

lib/application/widgets/theme/custom_text_field.dart
ggirotto: They have the exactly same visual behavior, but `border` has some issues when animating. I can't give you more details because I lost a lot of time trying to debug and understand, but couldn't figure it out

lib/application/hooks/rich_text_field_controller_hook.dart
ggirotto: Added this differences docs in the `RichTextFieldController` class docs

matuella (approved):
Looking good!

PR #230, opened 2021-11-04
Adds `UpdateCollectionPage`

ggirotto:


lib/application/pages/home/collections/update/update_collection_page.dart
```diff
@@ -0,0 +1,164 @@
...
+import 'package:memo/application/pages/home/collections/update/update_collection_memos.dart';
+import 'package:memo/application/pages/home/collections/update/update_providers.dart';
+import 'package:memo/application/theme/theme_controller.dart';
+import 'package:memo/application/view-models/home/update_collection_vm.dart';
+import 'package:memo/application/widgets/theme/custom_button.dart';
+import 'package:memo/application/widgets/theme/retry_container.dart';
+import 'package:memo/application/widgets/theme/themed_container.dart';
+import 'package:memo/application/widgets/theme/themed_tab_bar.dart';
+
+enum Segment { details, memos }
```
matuella, line 15: too broad of a name

matuella: shouldn't be private?

lib/application/pages/home/collections/update/update_collection_page.dart
```diff
@@ -0,0 +1,164 @@
...
+
+class UpdateCollectionPage extends HookWidget {
+  @override
+  Widget build(BuildContext context) {
+    final vm = readUpdateCollectionVM(context);
```
matuella, line 20: we should watch for the vm itself, as reading in builds is a bad pattern.

lib/application/pages/home/collections/update/update_collection_page.dart
```diff
@@ -0,0 +1,164 @@
...
+        return strings.memos;
+    }
+  }
+}
+
+class _BottomActionContainer extends HookWidget {
+  const _BottomActionContainer({required this.selectedSegment, required this.onSegmentSwapRequested});
+
+  final Segment selectedSegment;
+  final Function(Segment segment) onSegmentSwapRequested;
```
matuella, line 90:
```suggestion
  final void Function(Segment segment) onSegmentSwapRequested;
```

lib/application/pages/home/collections/update/update_collection_page.dart
```diff
@@ -0,0 +1,164 @@
...
+      case Segment.details:
+        return strings.details;
+      case Segment.memos:
+        return strings.memos;
+    }
+  }
+}
+
+class _BottomActionContainer extends HookWidget {
+  const _BottomActionContainer({required this.selectedSegment, required this.onSegmentSwapRequested});
```
matuella, line 87: Why nest all of this behavior in these widgets? It makes things much complex to understand. I would simply expose these functionalities to the "consumers" or the parent widgets, so that they can do whatever they want with actions made here in the container, and then, with that, change this bottom container to be simply configurable to all use-cases we need here

lib/application/coordinator/routes.dart
```diff
@@ -87,3 +87,16 @@ class SettingsPath extends AppPath {
   @override
   String get formattedPath => '/$name';
 }
+
+//
+// Update Collection
+//
+class UpdateCollectionPath extends AppPath {
```
matuella, line 94: Isn't it odd to name it "update" but it's actually creation and update at the same time?

lib/application/coordinator/routes.dart
```diff
@@ -87,3 +87,16 @@ class SettingsPath extends AppPath {
...
+  UpdateCollectionPath({this.collectionId});
+  static const name = 'update_collection';
+
+  final String? collectionId;
```
matuella, line 98: Don't we already have all the collection metadata once we get to this route? Shouldn't we simply fetch the associated `collection/$id/memos` once we are actually in the route?

lib/application/view-models/home/update_collection_vm.dart
```diff
@@ -0,0 +1,130 @@
...
+    required this.description,
+    required this.memoMetadata,
+  });
+
+  factory CollectionMetadata.empty() => const CollectionMetadata(name: '', description: '', tags: [], memoMetadata: []);
+
+  final String name;
+  final List<String> tags;
+  final String description;
+  final List<MemoMetadata> memoMetadata;
```
matuella, line 78: i don't see why we should nest them. Having them composable is better

matuella: So that the state can have a Collection and a Memo respective metadata variables

lib/application/view-models/home/update_collection_vm.dart
```diff
@@ -0,0 +1,130 @@
...
+      // TODO: Call services to save the collection
+      print('Saving collection...');
+    } on BaseException catch (exception) {
+      state = UpdateCollectionFailedSaving(exception, metadata: loadedState.collectionMetadata);
+    }
+  }
+}
+
+@immutable
+class CollectionMetadata {
```
matuella, line 65: equatable

lib/application/view-models/home/update_collection_vm.dart
```diff
@@ -0,0 +1,130 @@
...
+  factory CollectionMetadata.empty() => const CollectionMetadata(name: '', description: '', tags: [], memoMetadata: []);
+
+  final String name;
+  final List<String> tags;
+  final String description;
+  final List<MemoMetadata> memoMetadata;
+}
+
+@immutable
+class MemoMetadata {
```
matuella, line 82: equatable

lib/application/view-models/home/update_collection_vm.dart
```diff
@@ -0,0 +1,130 @@
...
+
+class UpdateCollectionLoading extends UpdateCollectionState {}
+
+class UpdateCollectionLoaded extends UpdateCollectionState {
+  const UpdateCollectionLoaded({required this.collectionMetadata});
+
+  final CollectionMetadata collectionMetadata;
+
+  bool get hasDetails => true;
+  bool get hasMemos => false;
```
matuella, line 107: what's going on here?

lib/application/view-models/home/update_collection_vm.dart
```diff
@@ -0,0 +1,130 @@
...
+  const MemoMetadata({required this.question, required this.answer});
+
+  factory MemoMetadata.empty() => const MemoMetadata(question: '', answer: '');
+
+  final String question;
+  final String answer;
+}
+
+@immutable
+abstract class UpdateCollectionState extends Equatable {
```
matuella, line 92: We must have an intermediate state for saving, as we will lose all the previous state if not

lib/application/widgets/theme/retry_container.dart
```diff
@@ -0,0 +1,40 @@
...
+import 'package:memo/application/constants/exception_strings.dart';
+import 'package:memo/application/constants/strings.dart' as strings;
+import 'package:memo/application/widgets/theme/custom_button.dart';
+import 'package:memo/core/faults/exceptions/base_exception.dart';
+
+/// A container that wraps a localized [exception] with a button that, when pressed, calls [onRetry].
+///
+/// This widget is mostly used in scenarios where a page couldn't load its essential data, meaning that there is no
+/// alternative other than retry loading its contents.
+class ExceptionRetryContainer extends StatelessWidget {
```
matuella, line 12: Should be a separate PR, but ok.

lib/application/widgets/theme/retry_container.dart
```diff
@@ -0,0 +1,40 @@
...
+  const ExceptionRetryContainer({required this.onRetry, required this.exception});
+
+  final VoidCallback onRetry;
+  final BaseException exception;
+
+  @override
+  Widget build(BuildContext context) {
+    final textTheme = Theme.of(context).textTheme;
+
+    return Column(
```
matuella, line 22: Will this fill the whole screen?

lib/application/widgets/theme/retry_container.dart
matuella: Can't we make tests for this widget?

lib/application/pages/home/collections/update/update_collection_page.dart
ggirotto: > too broad of a name

Do you think? I mean, considering that it's private which means that it applies to this screen only, isn't clear that it references to the tab bar segment?

lib/application/pages/home/collections/update/update_collection_page.dart
ggirotto: Updated the name to `useUpdateCollectionVM` and added `useProvider` due to the riverpod version

lib/application/coordinator/routes.dart
ggirotto: That's the pattern we've followed in other projects to name things that were creation and edition at the same time. We could also use `Set` but it seems stranger than `update` to me.

lib/application/view-models/home/update_collection_vm.dart
ggirotto: Added description and logic

lib/application/widgets/theme/retry_container.dart
ggirotto: <details>
<summary>Screenshot</summary>



</details>

lib/application/view-models/home/update_collection_vm.dart
ggirotto: Ideally this state is `UpdateCollectionLoaded` which may be constantly mutated with the user inputs

lib/application/pages/home/collections/update/update_collection_page.dart
ggirotto: I've split in different Widgets because the action button may assume different states based on the selected segment and the information that the user added. All use-cases are the following:

- When `Details` segment is selected the button is:
  - Enabled with `Create Collection` title when all the details information is complete and it has at least one memo;
  - Enabled with `Próximo` title when all details information is complete but the user doesn't have at least one memo;
  - Disabled with `Próximo` title when it doesn't have details information complete

- When `Memos` segment is selected the button is:
  - Enabled with `Create Collection` when `details` segment is complete and the user has at least one memo;
  - Disabled with `Create Collection` when there's missing information in `details` or the user doesn't have any memo created. 


Based on all these use-cases, I decided to create distinct buttons for each segment, since it's simpler to understand this division. If we move this use-cases scenarios to a single Widget, we would face an `if hell` scenario with a lot of ternary operations due to the different titles and actions.

lib/application/coordinator/routes.dart
ggirotto: We have no use-case that moves entities between routes. Do you think that we should introduce `transients` here? I mean, this route would receive an optional `transient` entity named `CollectionMetada`

lib/application/pages/home/collections/update/update_collection_page.dart
matuella: Yeah, it will pollute the global namespace with "Segment"

lib/application/widgets/theme/retry_container.dart
matuella: Shouldn't it be centered and fill all the available height?

lib/application/pages/home/collections/update/update_collection_page.dart
ggirotto: But it's private now. It's only visible inside the page

lib/application/widgets/theme/retry_container.dart
ggirotto: I'll update the component parent, because it's centered but the Column is not expanded in the parent

lib/application/view-models/home/update_collection_vm.dart
ggirotto: I misunderstood your comment at first. Added `UpdateCollectionSaving` and also `UpdateCollectionSaved`

lib/application/widgets/theme/exception_retry_container.dart
```diff
@@ -0,0 +1,40 @@
...
+/// This widget is mostly used in scenarios where a page couldn't load its essential data, meaning that there is no
+/// alternative other than retry loading its contents.
+class ExceptionRetryContainer extends StatelessWidget {
+  const ExceptionRetryContainer({required this.onRetry, required this.exception});
+
+  final VoidCallback onRetry;
+  final BaseException exception;
+
+  @override
+  Widget build(BuildContext context) {
```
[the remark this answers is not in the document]
ggirotto: Now it's centered because I've added `Center` wrapper around it in the parent (`UpdateCollectionPage`) widget

lib/application/view-models/home/update_collection_vm.dart
```diff
@@ -0,0 +1,168 @@
...
+      state = UpdateCollectionFailedLoading(exception);
+    }
+  }
+
+  @override
+  Future<void> saveCollection() async {
+    final loadedState = state as UpdateCollectionLoaded;
+    try {
+      // TODO: Call services to save the collection
+      print('Saving collection...');
```
matuella, line 61:
```suggestion
    final loadedState = state as UpdateCollectionLoaded;
    state = UpdateCollectionSaving...
    
    try {
      await Future<void>.delayed(const Duration(seconds: 2));
      // TODO: Call services to save the collection
      state = UpdateCollectionSaved...
```

lib/application/pages/home/collections/update/update_providers.dart
```diff
@@ -0,0 +1,18 @@
...
+final updateCollectionId = Provider<String?>((_) => throw UnimplementedError(), name: 'updateCollectionId');
+
+/// Syntax sugar for listening to [UpdateCollectionState] state updates.
+UpdateCollectionState useUpdateCollectionState(WidgetRef ref) {
+  final collectionId = ref.watch(updateCollectionId);
+  return ref.watch(updateCollectionVM(collectionId));
+}
+
+/// Syntax sugar for `ref.watch` the current [UpdateCollectionVM] provider.
+UpdateCollectionVM useUpdateCollectionVM(WidgetRef ref) {
```
matuella, line 15:
```suggestion
UpdateCollectionState watchUpdateCollectionState(WidgetRef ref) {
  final collectionId = ref.watch(updateCollectionId);
  return ref.watch(updateCollectionVM(collectionId));
}

/// Syntax sugar for `ref.watch` the current [UpdateCollectionVM] provider.
UpdateCollectionVM watchUpdateCollectionVM(WidgetRef ref) {
```

Can't we also make this a non-family provider? wdyt?

lib/application/pages/home/collections/update/update_collection_page.dart
```diff
@@ -0,0 +1,166 @@
...
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final state = useUpdateCollectionState(ref);
+
+    if (state is! UpdateCollectionLoaded) {
+      return const Center(child: CircularProgressIndicator());
+    }
+
+    final vm = useUpdateCollectionVM(ref);
+    final canSave = state.hasDetails && state.hasMemos;
```
matuella, line 159: would expose this "canSave" directly from the vm, instead of assuming it from the widget (UI)

lib/application/coordinator/routes.dart
matuella: That sure is a good question, I'm not sure. This is the same problem that we are facing in multiple other projects, which is: how do we handle "passing data" around multiple pages/states. I don't have a good enough answer now, so I think we could move on with the current solution.

lib/application/view-models/home/update_collection_vm.dart
```diff
@@ -0,0 +1,168 @@
...
+    try {
+      // TODO: Call services to load [collectionId] if necessary or create an empty collection metadata
+      await Future<void>.delayed(const Duration(seconds: 2));
+
+      // state = UpdateCollectionLoaded(collectionMetadata: CollectionMetadata.empty());
+      state = UpdateCollectionFailedSaving(
+        UrlException.failedToOpen(),
+        metadata: CollectionMetadata.empty(),
+        memosMetadata: const [],
+      );
```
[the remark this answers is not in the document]
ggirotto: Outdated I guess? Also, I would ignore this state management until we have the services good to go, since I'll constantly update it to error states to test the interface

lib/application/view-models/home/update_collection_vm.dart
ggirotto: Same as https://github.com/olmps/memo/pull/230#discussion_r753225167

lib/application/pages/home/collections/update/update_providers.dart
ggirotto: I made this changes in a [future PR](https://github.com/olmps/memo/pull/237/files#diff-35ff452a74be7334307fbfa02595615c1d02eec30fc17b6158156376324ce851R1-R10) after the discussion we had in Discord. Please ignore this family providers and functions for now

test/utils/fakes.dart
```diff
@@ -1,6 +1,12 @@
...
 final question = [
   {'question': '?'}
 ];
 final answer = [
   {'answer': '!'}
 ];
+
+class MockCallbackFunction extends Mock {
+  void call();
+}
```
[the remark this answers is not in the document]
ggirotto: Created a dedicated file

lib/application/pages/home/collections/update/update_collection_page.dart
ggirotto: You mean directly from the state, right? Created a getter there

test/utils/fakes.dart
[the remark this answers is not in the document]
ggirotto: Moved to the proper file

lib/application/pages/home/collections/update/update_providers.dart
ggirotto: I removed this auxiliary functions and the `family` providers in a [future PR](https://github.com/olmps/memo/pull/237/files#diff-35ff452a74be7334307fbfa02595615c1d02eec30fc17b6158156376324ce851R1-R10). Please ignore them by now

lib/application/coordinator/routes.dart
ggirotto: By current solution you mean using just the id or using the transient?

matuella (approved):
> I would ignore the state management in this function until we have the services good to go, since I'll constantly update it to error states and add unnecessary async Futures to test the interface and I may commit this changes accidentally. What do you think?

That's fine by me, my only concern with those is that we can eventually "introduce" future inconsistencies if not letting those "mocked flows" to a minimal (yet loyal to the expected design). Nonetheless, it's pretty minimal and just like I've said, I'm fine with that.

PR #232, opened 2021-11-12
Memo Update Component

ggirotto:


lib/application/widgets/theme/memo_terminal.dart
```diff
@@ -0,0 +1,124 @@
...
+import 'package:memo/application/theme/theme_controller.dart';
+import 'package:memo/application/utils/bottom_sheet.dart';
+import 'package:memo/application/widgets/theme/custom_button.dart';
+import 'package:memo/application/widgets/theme/rich_text_field.dart';
+import 'package:memo/application/widgets/theme/terminal_window.dart';
+
+/// A terminal-styled component that presents a `Memo` question and answer that can be updated.
+///
+/// Use [questionController] and [answerController] to control the Memo content being edited.
+class MemoTerminal extends HookConsumerWidget {
```
matuella, line 18: This really look super-specific to the needs of your UI in the "Update Collection", as it's creeping with a lot of specific functionality, like "memoIndex" and "onRemove".

I would suggest to make it more specific to your use case and not to try and abstract it before needing it (because it's not really adaptable to be used in multiple places, cause it's filled with context-specific properties and behaviors).

matuella: `TerminalWindow` does indeed make sense and it's really independent, but this widget doesn't seems to me like the case.

lib/application/widgets/theme/memo_terminal.dart
```diff
@@ -0,0 +1,124 @@
...
+        context.verticalBox(Spacing.small),
+        RichTextField(
+          modalTitle: answerTitle,
+          placeholder: strings.updateMemoAnswerPlaceholder,
+          controller: answerController,
+        ),
+      ],
+    );
+
+    Future<void> removeDialogConfirmation() => showDestructiveOperationModalBottomSheet(
```
matuella, line 82: can we make it a method instead?

lib/application/widgets/theme/memo_terminal.dart
```diff
@@ -0,0 +1,124 @@
...
+          context,
+          ref,
+          title: strings.removeMemoTitle,
+          message: strings.removeMemoMessage,
+          destructiveActionTitle: strings.remove.toUpperCase(),
+          cancelActionTitle: strings.cancel.toUpperCase(),
+          onDestructiveTapped: () {
+            onRemove!();
```
matuella, line 90:
```suggestion
            // We can bang-operator the `onRemove` cause it's only called when the button is available, which is also determined by this same property.
            onRemove!();
```

lib/application/widgets/theme/memo_terminal.dart
```diff
@@ -0,0 +1,124 @@
...
+    this.scrollController,
+  });
+
+  /// The index of the current memo in the `Collection` `Memo`'s list.
+  final int memoIndex;
+
+  /// Controls the `Memo` question content.
+  final RichTextFieldController? questionController;
+
+  /// Controls the `Memo` answer content.
```
matuella, line 33: not sure if really useful

test/application/widgets/theme/memo_terminal_test.dart
```diff
@@ -0,0 +1,67 @@
...
+import 'package:memo/application/widgets/theme/custom_button.dart';
+import 'package:memo/application/widgets/theme/destructive_button.dart';
+import 'package:memo/application/widgets/theme/memo_terminal.dart';
+import 'package:memo/application/widgets/theme/rich_text_field.dart';
+import 'package:mocktail/mocktail.dart';
+
+import '../../../utils/fakes.dart';
+import '../../../utils/widget_pump.dart';
+
+void main() {
```
matuella, line 13: IMO, could be tested, up to you tho:

- `onRemove` is not called when tapping the widget and it's `null`.
- "private" controllers are created for those that aren't provided.

matuella (changes requested):
Be careful while creating strings, they usually have a prefix that is related with their own context, you can see all the examples created up until now.

lib/application/constants/strings.dart
```diff
@@ -15,6 +15,7 @@ import 'package:memo/domain/enums/resource_type.dart';
 //
 const oops = 'Oops';
 const tryAgain = 'Tente Novamente';
+const cancel = 'Cancelar';
```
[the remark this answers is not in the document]
ggirotto: Moved to `Application-wide strings`

lib/application/constants/strings.dart
```diff
@@ -71,6 +72,19 @@ const newCollection = 'Nova Coleção';
...
+const collectionName = 'Nome da Coleção';
+const collectionDescription = 'Descrição da Coleção';
+
+String updateMemoQuestionTitle(int memoIndex) => '### Pergunta $memoIndex';
+const updateMemoQuestionPlaceholder = 'Digite a questão';
+
+const updateMemoAnswer = '### Resposta';
+const updateMemoAnswerPlaceholder = 'Digite a resposta';
+
+const remove = 'Remover';
```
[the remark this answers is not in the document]
ggirotto: Moved to `Application-wide strings`

lib/application/widgets/theme/memo_terminal.dart
ggirotto: Removed

lib/application/widgets/theme/memo_terminal.dart
ggirotto: I wasn't aiming for abstraction, but I agree with your point. What about rename to `UpdateMemoTerminal`?

test/application/widgets/theme/memo_terminal_test.dart
ggirotto: > onRemove is not called when tapping the widget and it's null.

I'm not sure what you suggested here. What do you mean with `and it's null.`?

> "private" controllers are created for those that aren't provided.

I removed the controllers initialization because they're not used in `MemoTerminal`, although the test that you suggested would fit will in `CustomTextField`, `RichTextField` and `TagsField`. What do you think?

lib/application/widgets/theme/update_memo_terminal.dart
```diff
@@ -0,0 +1,121 @@
+import 'package:flutter/material.dart';
```
matuella, line 1: > I wasn't aiming for abstraction, but I agree with your point. What about rename to UpdateMemoTerminal?

Sure, but I would move it to the respective folder, as it doesn't make sense to be used other in that particular context.

matuella (approved):
Would just move the widget to the update-page-related folder

PR #236, opened 2021-11-14
Adds Memo Creation CTA Empty State

ggirotto:


lib/application/pages/home/collections/update/update_collection_memos.dart
```diff
@@ -1,8 +1,107 @@
...
+import 'package:memo/application/constants/strings.dart' as strings;
+import 'package:memo/application/theme/theme_controller.dart';
 
 class UpdateCollectionMemos extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
-    return Container(color: Colors.blue, child: Center(child: Text('Memos')));
+    return _CreateMemoEmptyState(
+      onTap: () {
+        print("teste");
```
matuella, line 14: Would put a TODO

lib/application/pages/home/collections/update/update_collection_memos.dart
```diff
@@ -1,8 +1,107 @@
...
+      },
+    );
   }
 }
+
+/// An empty state CTA to add a new `Memo` to a `Collection`.
```
matuella, line 20:
```suggestion
/// An empty state call-to-action to add a new `Memo` to its `Collection`.
```

lib/application/pages/home/collections/update/update_collection_memos.dart
```diff
@@ -1,8 +1,107 @@
...
+  const _CreateMemoEmptyState({this.onTap});
+
+  final VoidCallback? onTap;
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final theme = ref.watch(themeController);
+    final textTheme = Theme.of(context).textTheme;
+
+    return Material(
```
matuella, line 31: Maybe we could create one or two smaller widgets (could definitely be variables), but it would make this widget more readable.

lib/application/pages/home/collections/update/update_collection_memos.dart
```diff
@@ -1,8 +1,107 @@
...
+              ],
+            ),
+          ),
+        ),
+      ),
+    );
+  }
+}
+
+class _CreateButtonPainter extends CustomPainter {
```
matuella, line 66: Why do we have a custom (really custom, given that it uses a painter) for this specific flow?

ggirotto: The goal of this painter was to draw the + CTA (See SS below). Doing it with a painter gives us the flexibility to customize the + and the background circle colors. I sent a message to Bruno to see if we can make the + in the middle transparent to it use the background color. This way I think we may be able to colorize the background circle using `Asset.color`.

<details>
<summary>Screenshot</summary>

CleanShot 2021-11-21 at 15 15 31@2x

</details>

ggirotto: Replaced by an asset

matuella:
> The goal of this painter was to draw the + CTA (See SS below). Doing it with a painter gives us the flexibility to customize the + and the background circle colors. I sent a message to Bruno to see if we can make the + in the middle transparent to it use the background color. This way I think we may be able to colorize the background circle using Asset.color.

Sure, my point wasn't exactly questioning your implementation, but as to why we have to introduce another brand new button-style "behavior" only to fulfill this screen's empty state. As you've implemented today, it's fine by me, but it's certainly strange to introduce a new "set of themed-buttons" without having them properly documented as a proper application-wide widget that will be used for particular scenarios.

matuella (approved):

PR #237, opened 2021-11-17
Adds support to update collection details

ggirotto:
In this PR I suggest the usage of `StateNotifierProviders` to control the workflow of a form - in this use-case, the collection details. I honestly see no reasons in this case for using `rxdart` as we've discussed, even because we could use normal `StreamController`s if you disagree about this suggestion.

Open to discussion.


lib/core/faults/exceptions/validation_exception.dart
```diff
@@ -0,0 +1,17 @@
...
+      : amount = null,
+        super(type: ExceptionType.emptyField);
+
+  /// Thrown when the user exceeded the characters amount for a given field.
+  ValidationException.fieldLengthExceeded(int maxLength)
+      : amount = maxLength,
+        super(type: ExceptionType.fieldLengthExceeded);
+
+  /// Helps to provide a more accurate error description for a validation exception.
+  final int? amount;
```
ggirotto, line 16: Not sure about this one... Open to suggestions or ideas

matuella: Yeah, this is leaking to all `ValidationException`, sure not a good solution

lib/application/pages/home/collections/update/update_collection_page.dart
```diff
@@ -124,38 +153,29 @@ class _DetailsActionButton extends ConsumerWidget {
...
-
     return PrimaryElevatedButton(onPressed: state.hasDetails ? onPressed : null, text: buttonTitle.toUpperCase());
   }
 }
 
 class _MemosActionButton extends ConsumerWidget {
   @override
   Widget build(BuildContext context, WidgetRef ref) {
-    final state = useUpdateCollectionState(ref);
+    final vm = ref.read(updateCollectionVM.notifier);
```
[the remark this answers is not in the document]
matuella: Screen Shot 2021-11-19 at 12 10 58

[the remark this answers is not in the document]
matuella: https://riverpod.dev/docs/concepts/reading

lib/application/pages/home/collections/update/update_collection_details.dart
```diff
@@ -1,8 +1,127 @@
...
+    return const TagsField();
+  }
+}
+
+class _DescriptionField extends HookConsumerWidget {
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final theme = ref.watch(themeController);
+    final textTheme = Theme.of(context).textTheme;
+    final vm = ref.read(updateCollectionDetailsVM.notifier);
```
matuella, line 92: Ditto about reading in build

lib/application/pages/home/collections/update/update_collection_details.dart
```diff
@@ -1,8 +1,127 @@
...
+    final descriptionLength = state.metadata.description.plainText.length;
+    return RichTextField(
+      controller: controller,
+      focus: focus,
+      modalTitle: Text(
+        strings.detailsDescription,
+        style: textTheme.bodyText1?.copyWith(color: theme.primarySwatch.shade400),
+      ),
+      placeholder: strings.collectionDescription,
+      helperText: '$descriptionLength/${validators.collectionDescriptionMaxLength} caracteres',
```
matuella, line 121: leaking string

lib/application/pages/home/collections/update/update_collection_details.dart
```diff
@@ -1,8 +1,127 @@
...
+
+    final nameLength = state.metadata.name.length;
+    return CustomTextField(
+      controller: controller,
+      focusNode: focus,
+      labelText: strings.collectionName,
+      inputFormatters: [
+        LengthLimitingTextInputFormatter(validators.collectionNameMaxLength),
+      ],
+      helperText: '$nameLength/${validators.collectionNameMaxLength} caracteres',
```
matuella, line 61: leaking string

lib/application/view-models/home/update_collection_details_vm.dart
```diff
@@ -0,0 +1,88 @@
...
+import 'package:memo/core/faults/exceptions/base_exception.dart';
+import 'package:memo/domain/validators/collection_validators.dart';
+
+final updateCollectionDetailsVM = StateNotifierProvider.autoDispose<UpdateCollectionDetailsVM, UpdatedDetailsState>(
+  (ref) => UpdateCollectionDetailsVMImpl(metadata: ref.read(updateDetailsMetadata)),
+  dependencies: [updateDetailsMetadata],
+  name: 'updateCollectionDetailsVM',
+);
+
+abstract class UpdateCollectionDetailsVM extends StateNotifier<UpdatedDetailsState> {
```
matuella, line 16: Wasn't it possible to create a VM for each of them?

lib/application/widgets/unfocus_detector.dart
```diff
@@ -0,0 +1,19 @@
+import 'package:flutter/material.dart';
+
+/// A wrapper Widget that request unfocus when tapping outside [child].
+class UnfocusDetector extends StatelessWidget {
```
matuella, line 4: I'm not sure if I follow the usage of this widget. Could you explain why this is needed? If it's an "unfocus" detector, how come it doesn't have a "onUnfocus"? Also, wouldn't it conflict with other "focus" elements in the `child` element?

If, for example, we have multiple children in this `child` and one of them is focused, won't tapping on another one that requests focus, unfocusing the one existing and focus on the next one? I know it may sound like the obvious, but this shouldn't be done by us, instead, it should be handled by Flutter and can potentially lead to "visual inconsistencies".

Why don't we handle this "unfocusing" as per widget-basis and let each implementation handle when it should unfocus its context?

matuella (commented):
Shouldn't all of your `useEffect` be called only once, thus, passing an empty array of keys to its call?

lib/core/faults/exceptions/validation_exception.dart
ggirotto: I've tried a different approach. Let me know what do you think.

lib/application/view-models/home/update_collection_vm.dart
```diff
@@ -27,32 +37,40 @@ abstract class UpdateCollectionVM extends StateNotifier<UpdateCollectionState> {
...
-        metadata: CollectionMetadata.empty(),
-        memosMetadata: const [],
-      );
+      state = UpdateCollectionLoaded(collectionMetadata: CollectionMetadata.empty(), memosMetadata: const []);
+
+      // state = UpdateCollectionFailedSaving(
+      //   UrlException.failedToOpen(),
+      //   metadata: CollectionMetadata.empty(),
+      //   memosMetadata: const [],
+      // );
```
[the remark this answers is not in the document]
ggirotto: I removed, but as I suggested in previous PRs I would ignore this state changes because I normally let them commented to make easy to test the changes. They will be addressed when connecting with the services stuff.

Since this entire branch will land on `main` only after the services are good to go, this "dirty" code will not be merged in `main`.

lib/application/pages/home/collections/update/update_collection_page.dart
```diff
@@ -58,19 +60,46 @@ class _UpdateCollectionContents extends ConsumerWidget {
...
+        case _Segment.memos:
+          return UpdateCollectionMemos();
+      }
+    }
+
+    throw InconsistentStateError.layout('Unsupported subtype (${state.runtimeType}) of `UpdateCollectionState`');
+  }
+}
+
+class _UpdateCollectionDetails extends ConsumerWidget {
```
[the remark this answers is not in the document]
ggirotto: Because `ref.watch(updateCollectionVM.notifier);` uses `updateDetailsMetadata` provider, which is injected in `UpdateCollectionDetails`. See SS below.

<details>
<summary>Screenshot</summary>

CleanShot 2021-11-21 at 15 51 38


</details>

What I may do is to use `Consumer` to remove this wrapper Widget, like:

Before:
```dart
if (state is UpdateCollectionLoaded) {
      switch (selectedSegment) {
        case _Segment.details:
          return ProviderScope(
            overrides: [
              updateDetailsMetadata.overrideWithValue(state.collectionMetadata),
            ],
            child: _UpdateCollectionDetails(),
          );
        case _Segment.memos:
          return UpdateCollectionMemos();
      }
    }
```

After:
```dart
if (state is UpdateCollectionLoaded) {
      switch (selectedSegment) {
        case _Segment.details:
          return ProviderScope(
            overrides: [
              updateDetailsMetadata.overrideWithValue(state.collectionMetadata),
            ],
            child: Consumer(
              builder: (context, ref, child) {
                ref.listen<UpdatedDetailsState>(
                  updateCollectionDetailsVM,
                  (_, state) => vm.updateMetadata(metadata: state.metadata),
                );

                return UpdateCollectionDetails();
              },
            ),
          );
        case _Segment.memos:
          return UpdateCollectionMemos();
      }
    }
```

What do you think?

lib/application/view-models/home/update_collection_details_vm.dart
ggirotto: I've made some tests creating an abstract VM that would be able to validate a single field from a generic type, this way we could easily instantiate a new VM for each field that we have. Would appreciate if you share your thoughts about it:


<details>
<summary>Field VM</summary>

```dart
abstract class FieldValidatorVM<T> extends StateNotifier<FieldValidatorState> {
  FieldValidatorVM({required this.validator, required FieldValidatorState state}) : super(state);

  final ValidationFunction validator;

  void validate(String updatedValue);
}

class FieldValidatorVMImpl<T> extends FieldValidatorVM<T> {
  FieldValidatorVMImpl({
    required ValidationFunction validator,
    required T value,
  }) : super(validator: validator, state: FieldValidatorState<T>(value: value));

  @override
  void validate(String updatedValue) {
    try {
      validator(updatedValue);
      state = state.copyWith(updatedValue: updatedValue);
    } on BaseException catch (exception) {
      state = state.copyForInvalid(exception);
    }
  }
}

class FieldValidatorState<T> extends Equatable {
  const FieldValidatorState({required this.value});

  final T value;

  FieldValidatorState copyWith({T? updatedValue}) => FieldValidatorState<T>(value: updatedValue ?? value);

  FieldValidatorInvalid copyForInvalid(BaseException exception) =>
      FieldValidatorInvalid<T>(value: value, exception: exception);

  @override
  List<Object?> get props => [value];
}

class FieldValidatorInvalid<T> extends FieldValidatorState<T> {
  const FieldValidatorInvalid({required T value, this.exception}) : super(value: value);

  final BaseException? exception;

  @override
  List<Object?> get props => super.props..add(exception);
}
```

</details>

And it could be used like:

<details>
<summary>Usage</summary>

```dart
final updateCollectionNameVM =
    StateNotifierProvider.family.autoDispose<FieldValidatorVM, FieldValidatorState<String>, ValidationFunction>(
  (ref, validator) => FieldValidatorVMImpl<String>(validator: validator, value: ref.read(updateDetailsMetadata).name),
  dependencies: [updateDetailsMetadata],
  name: 'updateCollectionNameVM',
);

final updateCollectionDescriptionVM = StateNotifierProvider.family
    .autoDispose<FieldValidatorVM, FieldValidatorState<RichTextEditingValue>, ValidationFunction>(
  (ref, validator) => FieldValidatorVMImpl<RichTextEditingValue>(
      validator: validator, value: ref.read(updateDetailsMetadata).description),
  dependencies: [updateDetailsMetadata],
  name: 'updateCollectionDescriptionVM',
);

final updateCollectionTagsVM =
    StateNotifierProvider.family.autoDispose<FieldValidatorVM, FieldValidatorState<List<String>>, ValidationFunction>(
  (ref, validator) =>
      FieldValidatorVMImpl<List<String>>(validator: validator, value: ref.read(updateDetailsMetadata).tags),
  dependencies: [updateDetailsMetadata],
  name: 'updateCollectionTagsVM',
);
```

</details>

But I thought that it was getting too complex to just integrate with these 3 fields, and I started to doubt from this solution. Besides that, I couldn't see any benefits on creating distinct but specific VMs for each field (like `UpdateCollectionNameVM`, `UpdateCollectionDescriptionVM` and `UpdateCollectionTagsVM`).

lib/application/widgets/unfocus_detector.dart
ggirotto: > Could you explain why this is needed?

I think this is an easy to use and must-have when we have any screen that the keyboard may appear, since it's possible to hide the keyboard tapping anywhere that doesn't have a TextField.

> If it's an "unfocus" detector, how come it doesn't have a "onUnfocus"?

Because we may use from Flutter singleton `FocusManager.instance`. The idea is not to expose this control to the parent Widget do something, it's just to hide the keyboard by removing the focus from the focused Widget (if it exists).

> Also, wouldn't it conflict with other "focus" elements in the child element?

It doesn't. I've made a few tests and Widgets like `TextField` defer the gestures to its child, which mean that this Widget gesture detector will not be used when tapping intractable elements. We may add some unit tests for these scenarios (TextField, buttons, scrollviews, etc.), although I think it's not necessary, since we already used Widgets like this one in other projects and we haven't faced problems.

> Why don't we handle this "unfocusing" as per widget-basis and let each implementation handle when it should unfocus its context?

As I mentioned above, the goal is not to provide a custom behavior to the parent Widget, since the use-case here will be always the same: The user want to dismiss the keyboard (i.e, unfocus some TextFiled) by tapping in non-intractable elements).

ggirotto:
> Shouldn't all of your useEffect be called only once, thus, passing an empty array of keys to its call?

Makes sense, there's no need to run them on every build

ggirotto:
@matuella Missed your thoughts about using `StateNotifierProvider` as the fields validation instead `StreamController` or even external libs as `rxdart`. Do you liked the proposal?

lib/application/view-models/home/update_collection_details_vm.dart
matuella: > Besides that, I couldn't see any benefits on creating distinct but specific VMs for each field (like UpdateCollectionNameVM, UpdateCollectionDescriptionVM and UpdateCollectionTagsVM).

To not branch in a "off-topic" discussion, the TLDR of the benefits could be explained by referencing SOLID - and here, specifically the SRP (and somewhat the ISP) principles.

---

The solution is looking good, but I would say we would have to think a little bit more on it, more specifically about the "family" usage and it being a `StateNotifier`. I feel like we could improve it and make it significantly less verbose. Anyways, again, not work for this PR, we could start working on a draft for this one if you want.

lib/application/view-models/home/update_collection_details_vm.dart
```diff
@@ -0,0 +1,88 @@
...
+  final CollectionMetadata metadata;
+
+  UpdatedDetailsState copyWith({String? name, List<String>? tags, RichTextEditingValue? description}) =>
+      UpdatedDetailsState(metadata: metadata.copyWith(name: name, tags: tags, description: description));
+
+  UpdateDetailsInvalid copyForInvalidName(BaseException exception) =>
+      UpdateDetailsInvalid(nameException: exception, metadata: metadata);
+
+  UpdateDetailsInvalid copyForInvalidDescription(BaseException exception) =>
+      UpdateDetailsInvalid(descriptionException: exception, metadata: metadata);
```
matuella, line 73: Wouldn't this lose the previous exception like invalid name?

lib/application/view-models/home/update_collection_details_vm.dart
```diff
@@ -0,0 +1,88 @@
...
+class UpdatedDetailsState extends Equatable {
+  const UpdatedDetailsState({required this.metadata});
+
+  final CollectionMetadata metadata;
+
+  UpdatedDetailsState copyWith({String? name, List<String>? tags, RichTextEditingValue? description}) =>
+      UpdatedDetailsState(metadata: metadata.copyWith(name: name, tags: tags, description: description));
+
+  UpdateDetailsInvalid copyForInvalidName(BaseException exception) =>
+      UpdateDetailsInvalid(nameException: exception, metadata: metadata);
```
matuella, line 70: Wouldn't this lose the previous exception like invalid description?

lib/application/widgets/unfocus_detector.dart
matuella: Then I would rename it, because if I'm not mistaken, Flutter uses `Detector` as a passive widget that provides gesture/behaviors detection. Maybe `AlwaysUnfocus`?

And also, I would explain when it should unfocus, as the docs seem to not address these. Things like: if the child doesn't respond for gestures, it will trigger the unfocus, otherwise it won't.

This widget could be more tested IMO, as it has a really significant responsibility and a lot of possible use-cases. My suggestion on these tests would be with a `child` that DO handle gestures and another that DO NOT handle. Could also possibly test with fields, those that trigger the keyboard appearance.

matuella (commented):
> @matuella Missed your thoughts about using StateNotifierProvider as the fields validation instead StreamController or even external libs as rxdart. Do you liked the proposal?

Sorry about that.

As of today, and correct me if I'm wrong, it "fundamentally" substitutes the `Stream` by creating copies of the state. My personal opinion:

pros:
- We trade the functionality that a Stream normally provides when we opt to implement it "manually", having no more stream builders and handling all of the "controllers".
- We pack them all (all fields) together, which is nice (at first glance) when we have to "group" all the validations in a single output (normally, a form that needs to have all of its fields in a valid state.

cons:
- Losing the Stream means that we will have a much more verbose "approach" when we deal with errors, because we will always have to create exception-specific "states" and also store/handle all of them (such as stacking those errors in a "general manner"). This will make things eventually harder to deal with (I've added two comments on such scenarios that I think are problematic with the "copy" approach), scaling this would be a nightmare for more than two or three "fields".
- We are rebuilding "everything" with a big state that aggregates everything "in the screen/page" which also goes against the SRP of a "view model", and now, the "packing all of these fields in a single unit" can easily become a future problem, because let's say that we have 15 fields of a collection, and we are rebuilding everything and creating copy functions that have to keep handling all of 15 `BaseException`, not only these properties.

The TLDR is that I think while it seems like a good approach for now, it's a solution that doesn't scale well, while also not performant (in the sense of having to reload only what is needed). Streams provides us with handling errors in a much more succinct/functional manner, which I feel that are more developer-friendly (less error prone), with only having to remember to dispose the stream controllers when finished.

lib/application/pages/home/collections/update/update_collection_page.dart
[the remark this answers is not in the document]
ggirotto: Ahh got it. Made the changes.

lib/application/widgets/unfocus_detector.dart
ggirotto: > Flutter uses Detector as a passive widget that provides gesture/behaviors detection

But isn't that what we are aiming to do here? I mean, detect the act of unfocusing a field based on a tap gesture. But I'm ok with changing the name if you prefer.

> And also, I would explain when it should unfocus, as the docs seem to not address these. Things like: if the child doesn't respond for gestures, it will trigger the unfocus, otherwise it won't.

Updated the docs.

> This widget could be more tested IMO

I was writing more tests to the Widget when I realized that the unique test cover all the cases that we expect. If you take a look at it, it covers both scenarios: Interacting with a focusable Widget (TextFiled) and with a non-focusable Widget (Container) and verify if the focus is managed accordingly. I was thinking about writing tests with buttons, scrollviews and others but it doesn't make sense because they are not focusable. The usage of this usage implies wrapping TextFields and other Widgets that receive the user input, and these are covered by this unique unit test.

ggirotto:
@matuella Updated to use `rxdart` and streams as we've discussed

lib/application/widgets/unfocus_detector.dart
matuella: > But isn't that what we are aiming to do here? I mean, detect the act of unfocusing a field based on a tap gesture. But I'm ok with changing the name if you prefer.

Yes, that's what we would do if we would want to detect something, but not to take actions upon our behalf. It doesn't only detect but also acts upon.

That is my perspective on why it would make sense to not use the same naming as `Detector`, but that's up to you honestly, just my 2 cents, maybe like `AbsorbPointer` or `IgnorePointer`, we could use `UnfocusPointer`?

matuella (approved):

lib/application/widgets/unfocus_detector.dart
ggirotto: I liked `UnfocusPointer`. Renamed the Widget

PR #238, opened 2021-11-18
Adds Memos List

ggirotto:


lib/application/pages/home/collections/update/update_collection_memos.dart
```diff
@@ -1,23 +1,154 @@
...
+        AssetIconButton(images.chevronLeftAsset, onPressed: onLeftTapped),
+        Text('$currentPage/$pagesAmount', style: textTheme.subtitle2),
+        AssetIconButton(images.chevronRightAsset, onPressed: onRightTapped),
+      ],
     );
   }
 }
 
-/// An empty state call-to-action to add a new `Memo` to its `Collection`.
+/// An empty state CTA to add a new `Memo` to a `Collection`.
```
matuella, line 151: this was possibly a rebase issue

lib/application/view-models/home/update_collection_vm.dart
```diff
@@ -129,10 +137,13 @@ class CollectionMetadata extends Equatable {
...
 
-  factory MemoMetadata.empty() => const MemoMetadata(question: '', answer: '');
+  factory MemoMetadata.empty() => const MemoMetadata(question: RichTextEditingValue(), answer: RichTextEditingValue());
+
+  MemoMetadata copyWith({RichTextEditingValue? question, RichTextEditingValue? answer}) =>
+      MemoMetadata(question: question ?? this.question, answer: answer ?? this.answer);
 
-  final String question;
-  final String answer;
+  final RichTextEditingValue question;
```
matuella, line 145: This is leaking a "UI" value into the VM

lib/application/pages/home/collections/update/update_collection_page.dart
```diff
@@ -71,17 +70,24 @@ class _UpdateCollectionContents extends ConsumerWidget {
     }
 
     if (state is UpdateCollectionLoaded) {
+      late Widget body;
```
matuella, line 73:
```suggestion
      final Widget body;
```

lib/application/pages/home/collections/update/update_collection_page.dart
```diff
@@ -71,17 +70,24 @@ class _UpdateCollectionContents extends ConsumerWidget {
...
-          );
+          body = UpdateCollectionDetails();
+          break;
         case _Segment.memos:
-          return UpdateCollectionMemos();
+          body = UpdateCollectionMemos();
+          break;
       }
+
+      return ProviderScope(
```
matuella, line 84: Why can't we provide the overrides to each specific scope that needs it? Like `updateDetailsMetadata` to `UpdateCollectionDetails` and `updateMemosMetadata` to `UpdateCollectionMemos`?

lib/application/pages/home/collections/update/update_collection_memos.dart
```diff
@@ -1,23 +1,154 @@
 import 'package:flutter/material.dart';
```
matuella, line 1: I'm really missing docs on these

lib/application/pages/home/collections/update/update_collection_memos.dart
```diff
@@ -1,23 +1,154 @@
...
+import 'package:memo/application/widgets/theme/rich_text_field.dart';
+
+class UpdateCollectionMemos extends HookConsumerWidget {
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final vm = ref.read(updateCollectionMemosVM.notifier);
+    final state = ref.watch(updateCollectionMemosVM);
+    final currentPageIndex = useState(0);
+
+    final controller = usePageController(viewportFraction: 0.95);
```
matuella, line 24: could we add this to a const value?

lib/application/pages/home/collections/update/update_collection_memos.dart
```diff
@@ -1,23 +1,154 @@
...
+    final parentVM = ref.watch(updateCollectionVM.notifier);
+    ref.listen<UpdateMemosState>(updateCollectionMemosVM, (_, state) => parentVM.updateMemos(memos: state.memos));
+
+    useEffect(() {
+      void onPageUpdate() => currentPageIndex.value = controller.page!.toInt();
+
+      controller.addListener(onPageUpdate);
+      return () => controller.removeListener(onPageUpdate);
+    });
```
matuella, line 33:
```suggestion
    }, []);
```

lib/application/pages/home/collections/update/update_collection_memos.dart
```diff
@@ -1,23 +1,154 @@
...
+    final vm = ref.read(updateCollectionMemosVM.notifier);
+    final state = ref.watch(updateCollectionMemosVM);
+    final currentPageIndex = useState(0);
+
+    final controller = usePageController(viewportFraction: 0.95);
+    final parentVM = ref.watch(updateCollectionVM.notifier);
+    ref.listen<UpdateMemosState>(updateCollectionMemosVM, (_, state) => parentVM.updateMemos(memos: state.memos));
+
+    useEffect(() {
+      void onPageUpdate() => currentPageIndex.value = controller.page!.toInt();
```
matuella, line 29: when can page be `null`?

lib/application/pages/home/collections/update/update_collection_memos.dart
```diff
@@ -1,23 +1,154 @@
...
+          vm.updateMemoAtIndex(pageIndex, metadata: metadata.copyWith(answer: answerController.value));
+
+      questionController.addListener(onQuestionUpdate);
+      answerController.addListener(onAnswerUpdate);
+
+      return () {
+        questionController.removeListener(onQuestionUpdate);
+        answerController.removeListener(onAnswerUpdate);
+      };
+    });
```
matuella, line 111:
```suggestion
    }, []);
```

lib/application/pages/home/collections/update/update_collection_memos.dart
ggirotto: Definitively

lib/application/pages/home/collections/update/update_collection_page.dart
ggirotto: Unfortunatelly not. Updating providers dynamically is not possible. See:

<details>
<summary>Code</summary>

```dart
switch (selectedSegment) {
        case _Segment.details:
          return ProviderScope(
            overrides: [
              updateDetailsMetadata.overrideWithValue(state.collectionMetadata),
            ],
            child: UpdateCollectionDetails(),
          );

          break;
        case _Segment.memos:
          return ProviderScope(
            overrides: [
              updateMemosMetadata.overrideWithValue(state.memosMetadata),
            ],
            child: UpdateCollectionMemos(),
          );
          break;
      }
```

</details>

<details>
<summary>Error</summary>

CleanShot 2021-11-25 at 14 03 31@2x

</details>

lib/application/pages/home/collections/update/update_collection_memos.dart
ggirotto: No because all controllers hooks are not constants

lib/application/pages/home/collections/update/update_collection_memos.dart
ggirotto: Based on the docs:

```dart
  /// There are circumstances that this [PageController] can't know the current
  /// page. Reading [page] will throw an [AssertionError] in the following cases:
  ///
  /// 1. No [PageView] is currently using this [PageController]. Once a
  /// [PageView] starts using this [PageController], the new [page]
  /// position will be derived:
  ///
  ///   * First, based on the attached [PageView]'s [BuildContext] and the
  ///     position saved at that context's [PageStorage] if [keepPage] is true.
  ///   * Second, from the [PageController]'s [initialPage].
  ///
  /// 2. More than one [PageView] using the same [PageController].
```

Basically is when the controller has no or more than one corresponding `PageView`. I believe we shouldn't worry about both cases, since this fore unwrap is being made in the controller listeners, which are triggered by interactions with the `PageView` directly.

lib/application/view-models/home/update_collection_vm.dart
ggirotto: I've added a new entity named `MemoUpdateContent`, which holds the rich and plain text versions of the question/answer. Let me know what do you think.

lib/application/pages/home/collections/update/update_collection_page.dart
matuella: Have you tried using different keys? The error seems to be that you're changing the `ProviderScope` overrides, which is not true, you're just sending a completely new child.

lib/application/pages/home/collections/update/update_collection_memos.dart
matuella: I meant the viewportFraction

lib/application/pages/home/collections/update/update_collection_memos.dart
ggirotto: I tried to put some more docs but I'm struggling about which docs would be useful here. I'm a little biassed by the page development, so everything looks obvious to me in this page

lib/application/pages/home/collections/update/update_collection_page.dart
ggirotto: But in the code above the overrides are changing, from this:

```dart
overrides: [
              updateDetailsMetadata.overrideWithValue(state.collectionMetadata),
            ]
```

to this

```dart
overrides: [
              updateMemosMetadata.overrideWithValue(state.memosMetadata),
            ],
```

lib/domain/transients/update_collection_metadata.dart
```diff
@@ -0,0 +1,67 @@
...
+  final MemoUpdateContent question;
+  final MemoUpdateContent answer;
+
+  @override
+  List<Object?> get props => [question, answer];
+}
+
+/// Holds `Memo` question/answer properties when updating a `Collection`.
+@immutable
+class MemoUpdateContent extends Equatable {
```
ggirotto, line 48: @matuella Created this transient `MemoUpdateContent` and moved the others MemoUpdateCollection` and `MemoUpdateMetadata` to here as well

matuella: Does it make sense to move them to domain/transient, if they only belong to that particular UI's view model?

lib/application/pages/home/collections/update/update_collection_page.dart
matuella: up

lib/application/pages/home/collections/update/update_collection_memos.dart
```diff
@@ -1,18 +1,168 @@
...
+    return UpdateMemoTerminal(
+      memoIndex: pageIndex + 1,
+      questionController: questionController,
+      answerController: answerController,
+      onRemove: onRemove,
+    );
+  }
+}
+
+/// Provides left-right navigation between `Memo`s from a `Collection`.
```
matuella, line 140:
```suggestion
/// Directional actions between all [_MemoPage] belonging to a horizontal [PageView].
```

lib/application/pages/home/collections/update/update_collection_memos.dart
```diff
@@ -1,18 +1,168 @@
...
+class _NavigationIndicator extends StatelessWidget {
+  const _NavigationIndicator({
+    required this.currentPage,
+    required this.pagesAmount,
+    this.onLeftTapped,
+    this.onRightTapped,
+  });
+
+  final int currentPage;
```
matuella, line 149:
```suggestion
  /// Visually indicates the current page's index.
  final int currentPage;
```

lib/application/pages/home/collections/update/update_collection_memos.dart
matuella: I'll suggest some.

lib/application/pages/home/collections/update/update_collection_memos.dart
matuella: up

lib/application/pages/home/collections/update/update_collection_page.dart
matuella: And that would be fine with a new `key` wouldn't it?

lib/application/pages/home/collections/update/update_collection_memos.dart
```diff
@@ -1,18 +1,168 @@
...
+  final int pagesAmount;
```
matuella, line 150:
```suggestion
  /// Displays the total available [_MemoPage].
  final int pagesAmount;
```

lib/application/pages/home/collections/update/update_collection_memos.dart
```diff
@@ -1,18 +1,168 @@
...
+          pagesAmount: state.memos.length + 1,
+          onLeftTapped: currentPageIndex.value > 0 ? onPreviousTapped : null,
+          onRightTapped: currentPageIndex.value < state.memos.length ? onNextTapped : null,
+        ),
+      ],
+    ).withSymmetricalPadding(context, vertical: Spacing.medium);
+  }
+}
+
+/// Presents editable `Memo` [metadata] wrapped around [UpdateMemoTerminal].
```
matuella, line 87:
```suggestion
/// Editable `Memo` [metadata] wrapped around an [UpdateMemoTerminal].
```

lib/application/pages/home/collections/update/update_collection_memos.dart
```diff
@@ -1,18 +1,168 @@
...
+class _MemoPage extends HookConsumerWidget {
+  const _MemoPage({required this.pageIndex, required this.metadata, this.onRemove, Key? key}) : super(key: key);
+
+  final int pageIndex;
```
matuella, line 91: Could add docs here on what this index means

lib/application/pages/home/collections/update/update_collection_page.dart
matuella: And that would be fine with a new `key` wouldn't it?

lib/application/pages/home/collections/update/update_collection_page.dart
ggirotto: Forgot about this one

lib/application/pages/home/collections/update/update_collection_page.dart
ggirotto: Using keys worked like a charm

lib/domain/transients/update_collection_metadata.dart
ggirotto: That's related to the questions that we've about what a transient model is. This entities are reused between different UI/View Models, so it deserves a dedicated file so all common files can import it (otherwise we would have a VM importing another VM just to get this entities).

We have defined as transients:

> Just like the models/, the transients represent a set of business structures, although there's a clear distinction between these and models - they exist only in memory, meaning that they make sense only in some particular contexts required by the application's lifecycle.

Following this definition, makes sense to this models be here IMO, what do you think?

lib/application/pages/home/collections/update/update_collection_page.dart
```diff
@@ -79,8 +78,14 @@ class _UpdateCollectionContents extends ConsumerWidget {
             ],
             child: UpdateCollectionDetails(),
           );
+
         case _Segment.memos:
-          return UpdateCollectionMemos();
+          return ProviderScope(
```
matuella, line 83: You said using keys worked, but I don't see them here, am I missing something?

lib/application/pages/home/collections/update/update_collection_memos.dart
matuella: Should the child know its position in the parent? Looks to me like it would be easier for the parent, that controls the list, to know where it's inserted and what it should do with a given child event.

lib/domain/transients/update_collection_metadata.dart
matuella: My only concern with this is that while "multiple" VMs use the same structure, it still belongs to this "update" flow, which could be considered a "single thing" in terms of application-wide structures.

The definition of "transients" are pretty weak right now and I think, like a couple of other architectural decisions, that we should revisit them, though it's tangential to this PR. Given this issue, I would suggest that we create a dedicated file for this structure in the same folder of this application/ domain category, which would be home/collections/update

matuella: suggestion though

lib/application/pages/home/collections/update/update_collection_page.dart
ggirotto: My bad, forgot to commit the changes

lib/application/pages/home/collections/update/update_collection_memos.dart
ggirotto: I've moved the responsibility of warning the VM about the changes to the parent, but we still have to have this index here because we visually show to the user what question is he seeing:

<details>
<summary>Screenshot</summary>

CleanShot 2021-11-29 at 11 17 55@2x

</details>

But now it only serve as a visual indicator, and it's not used anymore to update the VM

lib/domain/transients/update_collection_metadata.dart
ggirotto: Agreed

lib/application/pages/home/collections/update/update_collection_memos.dart
```diff
@@ -48,6 +48,7 @@ class UpdateCollectionMemos extends HookConsumerWidget {
             key: ValueKey(metadata),
             pageIndex: index,
             metadata: metadata,
+            onUpdate: (memoMetadata) => vm.updateMemoAtIndex(index, metadata: metadata),
```
ggirotto, line 51: This wrong reference was fixed in the next PR. Please ignore.

```suggestion
            onUpdate: (memoMetadata) => vm.updateMemoAtIndex(index, metadata: memoMetadata),
```

matuella (approved):
Thanks for the work

PR #241, opened 2021-11-24
Adds missing `tags` collection to `firestore-rules`

ggirotto:


ggirotto:
@matuella As we've discussed in Discord, I've updated this PR removing the guards and only added a missing collection to the rules

matuella (commented):
Why exactly is `tags` needed, now that we won't suggest it anymore? Can't it be simply an array?

ggirotto:
I forgot that we replaced this subcollection by an array. I'll close this PR then and update Miro data structure to reflect this decision.

PR #242, opened 2021-11-27
Adds Update Collections bash script and workflow

ggirotto:
Adds the typescript tool script that outputs the collections updates in the format

```typesciprt
{
  added: string[],
  updated: string[],
  removed: string[]
}
```

And adds the workflow that uses this output to run `sync-collections` script

### DIFF
--- a/.github/workflows/update-collections.yml
+++ b/.github/workflows/update-collections.yml
@@ -25,8 +25,8 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: ${{ env.node_version }}
-          # cache: "npm"
-          # cache-dependency-path: "**/package-lock.json"
+          cache: "npm"
+          cache-dependency-path: "**/package-lock.json"
 
       # Install project dependencies.
       - name: Install Firebase Dependencies
--- a/.github/workflows/update-collections.yml
+++ b/.github/workflows/update-collections.yml
@@ -1,0 +1,45 @@
+# Triggers on every release and triggers `sync-collections` scripts by sending the added, updated and removed
+# collections ids
+name: Update Collections
+
+on:
+  release:
+    types:
+      - published
+
+env:
+  node_version: "16.x"
+
+jobs:
+  update-collections:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./firebase
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      # Setup Node.js with version `env.node_version`.
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v2
+        with:
+          node-version: ${{ env.node_version }}
+          # cache: "npm"
+          # cache-dependency-path: "**/package-lock.json"
+
+      # Install project dependencies.
+      - name: Install Firebase Dependencies
+        run: npm i
+
+      # Retrieve the collections that were added, updated and removed.
+      #
+      # `sed -e '1,4d'` shell command removes the first 4 executions lines that output when running `npm run ...`.
+      - name: Get Collections Diff
+        run: |
+          collections_diff=$(npm run tools:collections-diff | sed -e '1,4d')
+          echo "COLLECTIONS_DIFF=$(echo $collections_diff)" >> $GITHUB_ENV
+
+      # Run script that sync the collections.
+      - name: Run sync-collections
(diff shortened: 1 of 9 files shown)


firebase/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "memo-firebase",
   "engines": {
-    "node": "14"
+    "node": "16"
```
ggirotto, line 4: Update Node version to latest stable version to use the same version we use in our development machines and also because node `14` has some issues with `postinstall` script, specially in CI's

firebase/package.json
```diff
@@ -42,6 +43,7 @@
     "prettier": "^2.4.x",
     "sinon": "^10.0.x",
     "ts-node": "^10.2.x",
+    "typescript": "^4.5.x",
```
ggirotto, line 46: Adds missing typescript dependency to the project (we have installed as a global package in our machines, but the CI hasn't)

firebase/tools/collections-diff.ts
```diff
@@ -0,0 +1,154 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
ggirotto, line 1: Not sure about this script folder. It is a script but it's not related with the application domain, it's related to the CI environment that we are using.

.github/workflows/update-collections.yml
```diff
@@ -0,0 +1,45 @@
...
+          # cache: "npm"
+          # cache-dependency-path: "**/package-lock.json"
+
+      # Install project dependencies.
+      - name: Install Firebase Dependencies
+        run: npm i
+
+      # Retrieve the collections that were added, updated and removed.
+      #
+      # `sed -e '1,4d'` shell command removes the first 4 executions lines that output when running `npm run ...`.
```
matuella, line 37: Isn't this really error-prone to npm changing stuff? How could we test this script in some way to guarantee its integrity?

.github/workflows/update-collections.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      # Setup Node.js with version `env.node_version`.
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v2
+        with:
+          node-version: ${{ env.node_version }}
+          # cache: "npm"
+          # cache-dependency-path: "**/package-lock.json"
```
matuella, line 29: any reason for these comments?

firebase/tools/collections-diff.ts
matuella: I think it is related to the application. Instead of passing the arguments directly to the sync use-case, maybe it could request to some gateway that knows how to make a diff? Not sure, there are a couple of ways to implement this, but just like we are accessing file system with the gateway, using the github API is another gateway. Just because we are consuming it through a CLI, it doesn't mean it's not a gateway. I would look into fetching it with a HTTP though, cause running CLI commands through a node application could be security-error-prone, which could be catastrophic, although if well built, I would say fine.

matuella (commented):
1. Just have to be aware if there is any issue with node 16 in firebase functions because it's still in beta.
2. I was thinking of running this script whenever a new diff makes sense in `main`. I think that attaching a github release with a collection update might not be a good solution, since we don't want necessarily to wait for a release to actually update the local/cloud collections. IMO this script should be ready to sync whenever there is a valid diff in `main`, and not attached to the release cycle.

.github/workflows/update-collections.yml
ggirotto: Sorry about that, I forgot to uncomment. I'm currently using [`act`](https://github.com/nektos/act) to test the actions locally instead testing in Github, and it has some specific limitations (NPM cache is one of them...)

ggirotto:
> 1. Just have to be aware if there is any issue with node 16 in firebase functions because it's still in beta.

I didn't knew that. I had some problems when using Node 14 but it may be the case that was a problem with `act` and not with Node. I'll do some tests in the Github environment instead in my local machine

.github/workflows/update-collections.yml
ggirotto: Just realized that `npm` has a `--silent` flag that can be used here.

ggirotto:
Closed in favor of #247, #248 and #249. Further discussions and changes will be handled there

PR #247, opened 2021-12-01
Adds `SyncCollections` faults, gateways and repos

ggirotto:


firebase/src/core/data/gateways/shell-gateway.ts
```diff
@@ -0,0 +1,20 @@
+import ShellError from "@faults/errors/shell-error";
+import { exec } from "child_process";
+
+/** Exposes access to perform shell operations. */
+export class ShellGateway {
+  /** Runs {@link command} in shell and returns an UTF-8 decoded string. */
+  async run(command: string): Promise<string> {
```
matuella, line 7: Shouldn't this return the exit status code of the script?

firebase/src/core/data/repositories/git-repository.ts
```diff
@@ -0,0 +1,31 @@
+import { ShellGateway } from "@data/gateways/shell-gateway";
+
+export class GitRepository {
```
matuella, line 3: Can we add tests for these?

firebase/src/core/data/gateways/shell-gateway.ts
```diff
@@ -0,0 +1,20 @@
+import ShellError from "@faults/errors/shell-error";
+import { exec } from "child_process";
+
+/** Exposes access to perform shell operations. */
+export class ShellGateway {
```
matuella, line 5: Can we add tests for these?

firebase/src/core/data/repositories/git-repository.ts
matuella: Maybe we could add docs for the class?

firebase/src/core/data/gateways/shell-gateway.ts
matuella: Maybe refer how this gateway uses node to open a shell process?

matuella (commented):
Really liked the approach! I just missed some tests

firebase/src/core/data/gateways/shell-gateway.ts
ggirotto: I don't think that returning the exit status make sense. We need to return the output because the gateway user may be interested in the shell output instead if it had success (exit 0) or not (exit 1). Also, the success of the command is measured by the Promise.resolve or Promise.reject, so I think we are good not using the exit status.

firebase/src/core/data/repositories/git-repository.ts
ggirotto: Added small docs. Let me know what do you think

firebase/src/core/data/repositories/git-repository.ts
```diff
@@ -0,0 +1,32 @@
+import { ShellGateway } from "@data/gateways/shell-gateway";
+
+/** Interfaces access to the project git repository information. */
```
matuella, line 3:
```suggestion
/** Git utilities applied to the current node process root folder. */
```

firebase/src/core/data/gateways/shell-gateway.ts
```diff
@@ -0,0 +1,24 @@
+import ShellError from "@faults/errors/shell-error";
+import { exec } from "child_process";
+
+/**
+ * Exposes access to perform shell operations.
+ *
+ * Uses `Node` to invoke shell processes.
```
matuella, line 7:
```suggestion
 * Uses node's {@link exec} to invoke shell processes.
```

nit

matuella (approved):

PR #248, opened 2021-12-01
Adds `DiffCollections` use-case and updates scripts

ggirotto:


firebase/package.json
```diff
@@ -6,6 +6,7 @@
   "private": true,
   "scripts": {
     "postinstall": "run-p install:*",
+    "setup-ci": "npm i && run-p install:*",
```
ggirotto, line 9: `postinstall` has some problems when running in sudo users, such as in Docker environments or in CI machines. An alternative would be to use `--unsafe-perm` flag, which is dangerous. So I've decided to create this extra script that does the setup for CI environments

firebase/package.json
```diff
@@ -21,9 +22,10 @@
     "build": "run-p --print-label build:*",
     "build:functions": "tsc -p tsconfig.functions.json",
     "functions:serve": "run-s --print-label build:functions functions:start-emulators watch",
-    "functions:fetch-env": "firebase functions:config:get > src/core/.runtimeconfig.json",
```
ggirotto, line 24: The env file must be placed in the root instead in the `core` folder, otherwise firebase emulator can't locate it

ggirotto:
@matuella Updated based on the discussion we had in Discord

firebase/src/core/domain/use-cases/collections/sync-collections.ts
```diff
@@ -34,20 +47,23 @@ export class SyncCollectionsUseCase {
    * Local collections are considered our source of truth (what we expect), meaning that any difference between what's
    * stored (what we have), should be respectively added, updated or removed.
    *
-   * The argument must specify a list of ids that were {@link addedOrUpdated} (no matter which operation, a set will be
-   * performed) and those that were {@link removed}.
+   * Communicates with Github to fetch a list of ids that were {@link addedOrUpdated} (no matter which operation, a set
```
matuella, line 50: links probably broken

firebase/src/core/domain/use-cases/collections/sync-collections.ts
```diff
@@ -127,4 +143,59 @@ export class SyncCollectionsUseCase {
...
     return [updateableMemos, deletableMemosIds];
   }
+
+  /**
+   * Returns the changed (added, updated and removed) collections ids.
+   *
+   * The collections diff is made by using `git ...` shell commands, so make sure to run the current use-case in the
+   * expected git HEAD.
+   */
+  async #collectionsDiff(): Promise<UpdatedCollections> {
```
matuella, line 153:
```suggestion
  async #diffCollections(): Promise<UpdatedCollections> {
```
following the other private function signature

matuella (approved):
LGTM, just some docs changes

PR #251, opened 2021-12-01
Updates project to only use `tsconfig-paths` when running tests and scripts

ggirotto:


matuella (approved):
Forgot to update these, as we aren't using any `functions`. Thanks for the changes.

Issue #223, opened 2021-10-28
Rich Text Editor


ggirotto:
### Summary

In favor of using [`flutter_quill`](https://github.com/singerdmx/flutter-quill) we will create our own rich text editor.

### Features

- The editor must support `readonly` mode.
- Must support `bold`, `italic`, `underline` and `code block` stylish properties.
- Must support interface customizations:
  - TextStyle
  - Background color
  - Cursor color
  - Placeholder
  - Placeholder text style
- The stylish properties must follow the same behavior as the most common editors: Update the selected text if it exists or apply the style to the new typed text otherwise.
- The keyboard type must be overridable.

PR #253, opened 2021-12-02
Memo schema tests

ggirotto:


firebase/package.json
```diff
@@ -48,6 +48,7 @@
     "@types/sinon": "^10.0.x",
     "@typescript-eslint/eslint-plugin": "^4.32.x",
     "@typescript-eslint/parser": "^4.32.x",
+    "ajv": "^8.8.2",
```
ggirotto, line 51: Moved to "base" package.json to be used in tests

firebase/src/core/domain/models/memo.ts
```diff
@@ -1,16 +1,36 @@
...
+    underline?: boolean;
+    codeBlock?: boolean;
+  };
+}
+
 export interface Memo {
-  readonly id: string;
-  readonly question: Record<string, unknown>[];
-  readonly answer: Record<string, unknown>[];
+  readonly uniqueId: string;
```
ggirotto, line 15: Just to point that I personally do prefer `id` over `uniqueId`, as I mentioned when we were developing Flutter client, but we need to create a standard. Or everything becomes `id` or `uniqueId`. We already have `uniqueId` set as the property in the flutter models and collections json's

firebase/src/core/domain/models/memo.ts
```diff
@@ -1,16 +1,36 @@
...
+  readonly question: MemoContent[];
+  readonly answer: MemoContent[];
 }
 
+const memoContentValidationSchema = Joi.object({
```
matuella, line 20:
```suggestion
const memoQuillValidationSchema = Joi.object({
```

Just to enforce where we are taking these from. I think being specific here is beneficial.

Maybe also add a succinct doc?

firebase/test/core/data/schemas/memo.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+          insert: "Question string",
+        },
+      ],
+      answer: [
+        {
+          insert: "Answer string",
+        },
+      ],
+    };
+  });
```
matuella, line 26: why not transform this into a `_newRawMemo()` function that returns the referred json? It gives a much more direct approach and don't have to manipulate a "wider-scope" variable

matuella (commented):
Tests looking good!

firebase/src/core/domain/models/memo.ts
ggirotto: But should we update the collections jsons and also Flutter model to use id as well? Otherwise we'll use `id` in the server and `uniqueId` in the client. I highlight what I've said back in the client development that `id` implicitly means that's an unique identifier IMO, that's no reason to emphasis the `unique` stuff.

PS: We would need this minor refactoring in a separate PR instead mixing with this one stuff

firebase/.vscode/launch.json
```diff
@@ -29,5 +29,14 @@
     //     "'${workspaceRoot}/node_modules/mocha/bin/_mocha -r ts-node/register -r tsconfig-paths/register --file \"${workspaceRoot}/test/firebase-rules/setup.ts\" --recursive \"${workspaceRoot}/test/firebase-rules/**/*.spec.ts\" --no-timeouts --exit'"
     //   ]
     // },
+    {
+      "name": "Debug current test file",
```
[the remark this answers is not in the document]
ggirotto: Better

firebase/src/core/domain/models/memo.ts
ggirotto: > Just to enforce where we are taking these from. I think being specific here is beneficial.

But aren't we going to follow the same format when we develop our own editor in https://github.com/olmps/memo/issues/223? If yes, I don't see why would we enforce `quill` in the name here, since we plan to remove it in the near future

firebase/test/core/data/schemas/memo.spec.ts
ggirotto: Agreed

matuella (commented):
> But should we update the collections jsons and also Flutter model to use id as well? Otherwise we'll use id in the server and uniqueId in the client. I highlight what I've said back in the client development that id implicitly means that's a unique identifier IMO, that's no reason to emphasis the unique stuff.
> 
> PS: We would need this minor refactoring in a separate PR instead mixing with this one stuff

I don't quite follow what you meant with "collections jsons", but yes, everything needs to be `id` since when we opted to update the strategy to use sub-collections of non-conflicting id's of these memos. The client will be eventually updated when the referencing models are mapped from firestore, as it's now mapping from flutter's assets folder + sembast local database.

firebase/src/core/domain/models/memo.ts
ggirotto: Renamed

matuella (approved):
Lgtm

PR #254, opened 2021-12-02
`Memo` entity validation tests

ggirotto:


firebase/test/core/domain/models/memo.spec.ts
```diff
@@ -0,0 +1,199 @@
+import ValidationError from "#faults/errors/validation-error";
```
ggirotto, line 1: I'm not sure about the tests strategy, which aims to simulate all possible scenarios based on the schema. We've a lot of invalid cases, but all of them may happen anytime, so I think all of them are necessary. It's quite big, but all test scenarios are very simple.

matuella (approved):

PR #256, opened 2021-12-03
`MemoRepository` tests

ggirotto:


firebase/test/.eslintrc
```diff
@@ -0,0 +1,5 @@
+{
```
ggirotto, line 1: Disabled all usages of `any` inside the `test/` directory exclusively, since it's a common sense that  we'll have a lot of mocks and any usages in the tests. Let me know what do you think of this decision.

firebase/test/core/data/gateways/shell-gateway.spec.ts
```diff
@@ -1,8 +1,8 @@
 import * as assert from "assert";
```
ggirotto, line 1: Fixes wrong imports after rebase

firebase/test/core/data/repositories/git-repository.spec.ts
```diff
@@ -1,7 +1,7 @@
 import * as assert from "assert";
```
ggirotto, line 1: Fixes wrong imports after rebase

firebase/test/core/data/repositories/memos-repository.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+import { MemosRepository } from "#data/repositories/memos-repository";
+import Ajv2020 from "ajv/dist/2020";
+import { SchemaValidator } from "#data/schemas/schema-validator";
+import { Memo } from "#domain/models/memo";
+import SerializationError from "#faults/errors/serialization-error";
+
+describe("MemosRepository", () => {
+  const firestoreMock = sinon.createStubInstance(FirestoreGateway);
+  const schemaMock = new SchemaValidator(new Ajv2020());
+  const memosRepo = new MemosRepository(<FirestoreGateway>(<any>firestoreMock), schemaMock);
```
ggirotto, line 13: I had to perform this type manipulation to TS don't complaint about the `sinon.Stub` type. I had absolutely no idea why it's complaining and I lost a lot of time trying to figure it out. If you have any idea about what's possible happening here, I'm open to suggestions.

matuella (changes requested):
Somewhat unrelated to the testing of this repository itself, but what would happen if we exceed all of these transactions with > than 500 operations, which is the firestore limit? I would guess our current solution doesn't support them, so I guess we must address this limitation (I would say if > 500 OPs, no transaction should be used), but again, offtopic to this, although a matter that we must probably address until finishing this server implementation.

firebase/test/core/data/repositories/memos-repository.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    it("should throw when one of the memos has invalid format", () => {
+      const malformedMemo: Memo = <any>{ id: "id2", foo: "bar" };
+      const mockMemosPerCollection = _newMemosPerCollection(mockCollectionId, [mockMemo]);
+      mockMemosPerCollection.set("collectionId2", [malformedMemo]);
+
+      assert.rejects(() => memosRepo.setMemos(mockMemosPerCollection), SerializationError);
+    });
+
+    it("should update a Memo using its raw representation", async () => {
```
matuella, line 45:
```suggestion
    it("should update a memo using its raw representation", async () => {
```

firebase/test/core/data/repositories/memos-repository.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  describe("removeMemosByIds", () => {
+    const mockCollectionId = "collectionId1";
+    const mockMemoId = "id1";
+    const memosIdsPerCollection = _newMemosIdsPerCollection(mockCollectionId, [mockMemoId]);
+
+    it("should remove all Memos from a collection inside a single transaction", async () => {
+      await memosRepo.removeMemosByIds(memosIdsPerCollection);
+
+      assert.ok(firestoreMock.runTransaction.calledOnce);
```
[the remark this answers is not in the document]
ggirotto: Sure

firebase/test/core/data/repositories/memos-repository.spec.ts
ggirotto: Awesome

ggirotto:
> Somewhat unrelated to the testing of this repository itself, but what would happen if we exceed all of these transactions with > than 500 operations, which is the firestore limit? I would guess our current solution doesn't support them, so I guess we must address this limitation (I would say if > 500 OPs, no transaction should be used), but again, offtopic to this, although a matter that we must probably address until finishing this server implementation.

You're right, but I'm not sure if we can do something related to this issue than placing docs in the `runTransaction` function. We cannot ensure how transactions are being made inside `context`.

firebase/test/core/data/repositories/memos-repository.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+import { MemosRepository } from "#data/repositories/memos-repository";
+import Ajv2020 from "ajv/dist/2020";
+import { SchemaValidator } from "#data/schemas/schema-validator";
+import { Memo } from "#domain/models/memo";
+import SerializationError from "#faults/errors/serialization-error";
+import createSinonStub from "#test/sinon-stub";
+
+describe("MemosRepository", () => {
+  const firestoreStub = createSinonStub(FirestoreGateway);
+  const schemaValidator = new SchemaValidator(new Ajv2020());
```
matuella, line 13: > This is not mocked/stubbed, we shouldn't rely on anything other than the unit being tested

Again, not sure why, but why isn't the schema being stubbed as well, if we are testing the repository?

firebase/test/core/data/repositories/memos-repository.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+
+function _newMemosPerCollection(collectionId: string, memos: Memo[]): Map<string, Memo[]> {
+  return new Map<string, Memo[]>([[collectionId, memos]]);
+}
+
+function _newMemosIdsPerCollection(collectionId: string, memosIds: string[]): Map<string, string[]> {
+  return new Map<string, string[]>([[collectionId, memosIds]]);
+}
+
+function _newRawMemo(props?: { id?: string; question?: string; answer?: string }): Memo {
```
matuella, line 122:
```suggestion
function newMemosPerCollection(collectionId: string, memos: Memo[]): Map<string, Memo[]> {
  return new Map<string, Memo[]>([[collectionId, memos]]);
}

function newMemosIdsPerCollection(collectionId: string, memosIds: string[]): Map<string, string[]> {
  return new Map<string, string[]>([[collectionId, memosIds]]);
}

function newRawMemo(props?: { id?: string; question?: string; answer?: string }): Memo {
```

firebase/test/core/data/repositories/memos-repository.spec.ts
ggirotto: You're right. Also added some asserts that the validation is being made when setting and getting the memos

matuella (approved):

PR #260, opened 2021-12-06
Adds `LocalPublicCollection` schema tests

ggirotto:


firebase/test/core/data/schemas/memo.spec.ts
```diff
@@ -5,88 +5,61 @@ import SerializationError from "#faults/errors/serialization-error";
```
ggirotto, line 5: Updated to follow the same format as the one proposed in the `local-public-collection` file (reduce visual complexity)

firebase/src/core/domain/models/collection.ts
```diff
@@ -5,10 +5,10 @@ import * as Joi from "joi";
 interface Collection {
   readonly id: string;
   readonly name: string;
+  readonly description: string;
+  readonly tags: string[];
+  readonly category: string;
 
-  readonly tags?: string[];
-  readonly category?: string;
-  readonly description?: string;
```
ggirotto, line 11: Updated these properties because, since we are talking about a public collection they should be required (at least the description should) IMO.

matuella: But then why is it required in the `Collection` interface? Public do indeed require ALL properties, but personal collections should not require anything other than an `id`, `name` and at least a single memo. We don't have a class representation for these in the server as we have no need for these up until now.

matuella: But then why is it required in the `Collection` interface? Public do indeed require ALL properties, but personal collections should not require anything other than an `id`, `name` and at least a single memo. We don't have a class representation for these in the server as we have no need for these up until now.

firebase/package.json
```diff
@@ -41,6 +41,7 @@
     "@typescript-eslint/eslint-plugin": "^4.32.x",
     "@typescript-eslint/parser": "^4.32.x",
     "ajv": "^8.8.2",
+    "ajv-formats": "^2.1.1",
```
ggirotto, line 44: Adds supporting for using `format: url` in Ajv schemaes

firebase/src/core/domain/models/collection.ts
ggirotto: I didn't get your point here, do you agree or disagree? `Public` collections should have such properties as required and `Private` not, but since we don't have private Collections in the server yet, this is not in this PR scope, right?

ggirotto:
@matuella Ready to be reviewed. This PR is greater than the original because I changed the strategy in the middle of the development. The following ones are smaller

firebase/src/core/data/schemas/entities-schemas/local-public-collection.json
```diff
@@ -9,6 +9,7 @@
     "memos": {
       "type": "array",
       "minItems": 1,
+      "uniqueItems": true,
```
matuella, line 12: Does `uniqueItems` work with objects? I mean, are properties order evaluated independently? Such as `{id:1,val:a} === {val:a,id:1}`?

firebase/test/core/data/schemas/local-public-collection.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+    }
+  });
+
+  it("should not throw when a optional property is not set", () => {
+    for (const optionalProperty of optionalProperties) {
+      const rawCollection = newRawLocalCollection();
+
+      delete rawCollection[optionalProperty];
+
+      doesNotThrow(() => validator.validateObject("local-public-collection", rawCollection), SerializationError);
```
matuella, line 40: does not throw with a `SerializationError`?

firebase/src/core/data/schemas/entities-schemas/local-public-collection.json
ggirotto: Yes. They use [`fast-deep-equal`](https://github.com/epoberezkin/fast-deep-equal) to perform the comparisons

firebase/test/core/data/schemas/local-public-collection.spec.ts
ggirotto: Removed

firebase/package.json
ggirotto: It seems to be contained in the `json-schema` [specification](https://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.7.3.6)

firebase/src/core/domain/models/collection.ts
ggirotto: We can take advantage from TS and override them to become required only in the `PublicCollection`. Updated the interfaces. Let me know what do you think.

matuella (commented):
Yep, that's really good!

I won't approve due to our discord discussion, as to abstract all these schemas/models tests. Let me know when you update them

firebase/package.json
```diff
@@ -40,12 +40,15 @@
...
     "@typescript-eslint/parser": "^4.32.x",
+    "@types/hapi__joi": "^17.1.7",
     "ajv": "^8.8.2",
+    "ajv-formats": "^2.1.1",
     "eslint": "^7.32.x",
     "eslint-config-prettier": "^8.3.x",
     "eslint-plugin-prettier": "^4.0.x",
     "firebase-admin": "^9.11.x",
     "firebase-tools": "^9.18.x",
+    "joi": "17.4.2",
```
ggirotto, line 51: This is a common-case for the suggestions I've made in https://github.com/olmps/memo/pull/255. This `joi` dependency must have the same version as the inner `Joi` dependency (src/core/package.json), otherwise the tests don't compile.

firebase/test/core/data/schemas/memo.spec.ts
```diff
@@ -2,153 +2,67 @@ import Ajv2020 from "ajv/dist/2020";
...
-      const rawMemo = _newRawMemo();
-
-      rawMemo.answer = true;
-
-      throws(() => validator.validateObject("memo", rawMemo), SerializationError);
-    });
+    validator.validate();
   });
 
   describe("Attributes - ", () => {
```
ggirotto, line 36: This `Attributes` could have been separated in a separate schema to also use the validator, but I think that it's not worth it to be honest, so I left this specific tests here

firebase/test/validator.ts
```diff
@@ -0,0 +1,259 @@
+import Ajv2020 from "ajv/dist/2020";
```
matuella, line 1: I think this implementation is looking good, although I have to add my concerns with using the term "validation/validate". I think what we should focus here is to be super specific about testing stuff. So, something like `entity-tester.ts`, as it implies that it's testing an entity and all of its properties.

As we use `Models` for (`domain/models`) objects and `Schema` (`data/schemas`) for raw data objects, I think it will fit nicely here, as we'd have `EntityTester` and the respective specializations named `SchemaTester` and `ModelTester`.

firebase/test/validator.ts
```diff
@@ -0,0 +1,259 @@
...
+   * Useful when validating enums.
+   */
+  readonly strictValue?: Map<string, string[]>;
+  /** Maps a string property to its maximum allowed length. */
+  readonly lengthRestricted?: Map<string, number>;
+  /** Maps a property to an incorrect type that shouldn't be acceptable. */
+  readonly incorrectTypes: Map<string, any>;
+}
+
+type ClassConstructor = () => any;
```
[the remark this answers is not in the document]
ggirotto: Renamed to `EntityBuilder`

firebase/test/validator.ts
```diff
@@ -0,0 +1,259 @@
...
+type ValidateFunction = (object: any) => void;
```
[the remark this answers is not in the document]
ggirotto: It becomes type-safer, but the goal here is to provide free-manipulation from the entity structure to the tests. As an example, we want to `delete` properties that couldn't be deleted if type was inferred, like `delete rawEntity[requiredProperty];`. We need to use `any` here to ensure this "free" manipulation in the entity to completely manipulate it to have unexpected states.

firebase/test/validator.ts
ggirotto: Agreed

matuella (approved):
Really nice structure to scale testing on entities and schemas!

PR #265, opened 2021-12-06
Adds `utils` tests

ggirotto:


firebase/test/core/utils/validate.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+    assert.throws(() => validate(objectSchema, object), ValidationError);
+  });
+
+  it("should add validation error details to ValidationError", () => {
+    const { error } = objectSchema.validate(object);
+    const errorDetails = error!.details.map((detail) => detail.message).join(",");
+    const expectedValidationError = new ValidationError({ message: errorDetails, origin: error });
+
+    assert.throws(() => validate(objectSchema, object), expectedValidationError);
+  });
```
ggirotto, line 24: Not sure about this one, but I think it's valid because it doesn't test the Joi implementation, but the `ValidationError` construction

firebase/package.json
```diff
@@ -35,6 +35,7 @@
   },
```
ggirotto, line 35: This is a good example that I've exemplified in our discussion in https://github.com/olmps/memo/pull/255

firebase/test/core/utils/deep-equal.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+    assert.ok(deepEqual);
+  });
+
+  it("should return false when two objects aren't strictly equal", () => {
+    const objectA = { a: { n: 0 } };
+    const objectB = { a: { n: "0" } };
+
+    const deepEqual = objectsEqual(objectA, objectB);
+
+    assert.ok(!deepEqual);
```
matuella, line 20: don't we have a NotOk or something?

matuella (approved):
Looking good

firebase/test/core/utils/deep-equal.spec.ts
ggirotto: Unfortunately not. The only ones that I've found are:

```typescript
            | 'equal'
            | 'notEqual'
            | 'deepEqual'
            | 'notDeepEqual'
            | 'ok'
            | 'strictEqual'
            | 'deepStrictEqual'
            | 'ifError'
            | 'strict'
```

PR #266, opened 2021-12-06
Validates if `memosOrder.length` equals to `memosAmount`

ggirotto:


firebase/test/core/domain/models/stored-public-collection.spec.ts
```diff
@@ -2,7 +2,7 @@ import ValidationError from "#faults/errors/validation-error";
 import { storedCollectionSchema } from "#domain/models/collection";
 import { doesNotThrow, throws } from "assert";
 import { validate } from "#utils/validate";
-import { newRawLocalCollection } from "#test/core/data/schemas/collections-fakes";
+import { newRawStoredCollection } from "#test/core/data/schemas/collections-fakes";
```
ggirotto, line 6: Also fixed `LocalCollection` usage instead `StoreLocation`

firebase/src/core/domain/models/collection.ts
```diff
@@ -84,12 +84,12 @@ export const localCollectionSchema = publicCollectionSchema.append({
...
-  memosOrder: Joi.array().items(Joi.string()).min(1).required(),
-});
-// TODO: How to compare between same-object properties?
-// .assert(Joi.ref("memosAmount"), Joi.ref("memosOrder.length"));
+export const storedCollectionSchema = publicCollectionSchema
+  .append({
+    memosAmount: Joi.number().integer().min(1).required(),
+    memosOrder: Joi.array().items(Joi.string()).unique().min(1).required(),
+  })
+  .assert(".memosAmount", Joi.ref("memosOrder.length"));
```
matuella, line 92: Oh that was tricky, really nice. Where did you find the notation for this usage? Maybe we could add a reference to the docs? I tried and didn't find anything.

matuella (approved):

firebase/src/core/domain/models/collection.ts
ggirotto: While searching where I found this solution, I found a better one here: https://github.com/sideway/joi/pull/1206.

Let me know what do you think

matuella (approved):

firebase/src/core/domain/models/collection.ts
ggirotto: Added docs

PR #267, opened 2021-12-07
Adds `EnvGateway` to initialize Firebase Admin based on Env

ggirotto:


firebase/src/core/data/gateways/firebase-app.ts
```diff
@@ -1,5 +1,9 @@
...
 
-// TODO: Properly initialize this, taking into consideration the environment
-process.env["FIRESTORE_EMULATOR_HOST"] = "localhost:8080";
-export const app = admin.initializeApp({ projectId: "test-rpoj" });
+export function initializeFirebase(options: { isLocalDevelopment: boolean }): admin.app.App {
+  if (options.isLocalDevelopment) {
+    process.env["FIRESTORE_EMULATOR_HOST"] = "localhost:8080";
+  }
+
+  return admin.initializeApp();
```
matuella, line 8: Does all emulators run properly if `projectId` is not used?

firebase/src/core/presentation/provider.ts
```diff
@@ -24,15 +26,18 @@ export default class Provider {
   }
 
   // eslint-disable-next-line @typescript-eslint/no-empty-function
-  private constructor() {}
+  private constructor() {
```
matuella, line 29: remove lint above

firebase/src/core/data/gateways/firebase-app.ts
ggirotto: You're right. I forgot that scripts depend from `projectId`. Updated.

firebase/src/core/data/gateways/firebase-app.ts
```diff
@@ -1,5 +1,18 @@
...
+  }
+
+  const serviceAccount = env.firebaseServiceAccount;
+  return admin.initializeApp({
+    projectId: serviceAccount.projectId,
+    credential: admin.credential.cert({
+      clientEmail: serviceAccount.clientEmail,
+      projectId: serviceAccount.projectId,
+      privateKey: serviceAccount.privateKey,
+    }),
```
ggirotto, line 16: The `credentials` initialization it's not required when running the `functions` environment. But when running scripts locally, we need to properly authenticate with Firebase. To support both scenarios, I suggest to we always initialize with Firebase Service Account metadata

matuella (approved):

PR #268, opened 2021-12-07
Adds type-checking to gateways tests

ggirotto:


firebase/test/core/data/gateways/filesystem-gateway.spec.ts
```diff
@@ -15,52 +15,50 @@ describe("FileSystemGateway", () => {
...
     await assert.rejects(() => fsGateway.readDirFilesAsStrings("any"), FilesystemError);
   });
 
   it("should reject when readFile throws", async () => {
     const fsGateway = new FileSystemGateway();
-    fsMock.expects("readdir").returns(fakeFiles);
-    fsMock.expects("readFile").rejects();
+    // Disable TS type-checking for `fsStub.readdir` call because `sinon` type definitions always invoke the last
+    // function signature by default when a function has overloads.
+    // See https://github.com/DefinitelyTyped/DefinitelyTyped/issues/36436 for more details about this issue.
```
matuella, line 42: Could we add this to `test/README` instead? I think it's really relevant and should not live only here next to this comment.

matuella (commented):
Nice improvements, sad to see that we have to deal with years-long issue that haven't been solved yet :/

firebase/test/core/data/gateways/filesystem-gateway.spec.ts
ggirotto: Sure

matuella (approved):

PR #273, opened 2021-12-08
Adds collections integrity unit tests

ggirotto:
Also adds updated - added `resources` and made other small adjustments - collections to `collections/` folder.

### DIFF
--- a/firebase/test/collections/integrity-check.spec.ts
+++ b/firebase/test/collections/integrity-check.spec.ts
@@ -1,16 +1,22 @@
 import * as assert from "assert";
 import * as fs from "fs";
-import Provider from "#presentation/provider";
+import Ajv2020 from "ajv/dist/2020";
+import { SchemaValidator } from "#data/schemas/schema-validator";
 
 describe("Collections Integrity Check", () => {
   let localCollectionsPaths: string[];
   let rawLocalCollections: any[];
 
   before(async () => {
-    localCollectionsPaths = await fs.promises.readdir(`${process.cwd()}/collections`);
+    const rootDir = process.cwd();
+    console.log(process.cwd());
+    localCollectionsPaths = await fs.promises.readdir(`${rootDir}/collections`);
 
-    const fsGateway = Provider.instance.fileSystemGateway;
-    const localCollections = await fsGateway.readDirFilesAsStrings("./collections");
+    const collectionsDir = `${rootDir}/collections`;
+    const files = await fs.promises.readdir(collectionsDir);
+    const localCollections = await Promise.all(
+      files.map((file) => fs.promises.readFile(`${collectionsDir}/${file}`, "utf-8"))
+    );
     rawLocalCollections = localCollections.map((collection) => JSON.parse(collection));
   });
 
@@ -34,7 +40,7 @@ describe("Collections Integrity Check", () => {
     });
   });
 
-  it("should have unique memo ids amongst themselves", () => {
+  it("should have unique memo ids in each collection", () => {
     const memosIds: string[] = [];
     for (const collection of rawLocalCollections) {
       const memos: any[] = collection["memos"];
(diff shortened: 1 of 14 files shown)


firebase/test/collections/integrity-check.spec.ts
```diff
@@ -1,6 +1,61 @@
...
+
+  before(async () => {
+    localCollectionsPaths = await fs.promises.readdir(`${process.cwd()}/collections`);
+
+    const fsGateway = Provider.instance.fileSystemGateway;
+    const localCollections = await fsGateway.readDirFilesAsStrings("./collections");
+    rawLocalCollections = localCollections.map((collection) => JSON.parse(collection));
+  });
+
+  it("should have at least one collection", () => assert.ok(rawLocalCollections.length > 0));
```
matuella, line 17: Why should it? I Don't think it's an useful test

firebase/test/collections/integrity-check.spec.ts
```diff
@@ -1,6 +1,61 @@
...
+  it("should have unique ids amongst themselves", () => {
+    const collectionIds: string[] = [];
+    rawLocalCollections.forEach((collection) => {
+      const collectionId: string = collection["id"];
+      assert.ok(!collectionIds.includes(collectionId), `Duplicate collection id "${collectionId}"`);
+      collectionIds.push(collectionId);
+    });
+  });
+
+  it("should have unique memo ids amongst themselves", () => {
```
matuella, line 39:
```suggestion
  it("should have unique memo ids in each collection", () => {
```

Why should they have a unique global id?

firebase/test/collections/integrity-check.spec.ts
```diff
@@ -1,6 +1,61 @@
...
+          !memosIds.includes(memoUniqueId),
+          `Duplicate memo id "${memoUniqueId}" in collection "${collection}"`
+        );
+        memosIds.push(memoUniqueId);
+      });
+    }
+  });
+
+  it("should have expected JSON structure", () => {
+    const schemaValidator = Provider.instance.schemaValidator;
```
matuella, line 55: it seems odd to use provider here, why not create an instance of the expected schema validator?

matuella (commented):
Shouldn't we move these tests to integration tests of some sort?

matuella:
Also, I've already removed the test collection in `https://github.com/olmps/memo/pull/276`. Will merge both #275 and #276 so you can rebase on that.

firebase/test/collections/integrity-check.spec.ts
ggirotto: I copied this test from the client, I though it was a definition, but we may remove this test

matuella (commented):
Will these collections properly run at the same time that the workflow is merged into `main`?

Like, we're developing in `server`, then, when merging into `main`, will the `.github/` changes work effectively with the script itself?

ggirotto:
> Will these collections properly run at the same time that the workflow is merged into `main`?
> 
> Like, we're developing in `server`, then, when merging into `main`, will the `.github/` changes work effectively with the script itself?

Good catch. Updated `update-collections` workflow to run the integrity tests before running the collections sync.

matuella:
> Good catch. Updated update-collections workflow to run the integrity tests before running the collections sync.

But does it change the fact that all collections are going to be merged before the workflow is available?

ggirotto:
> But does it change the fact that all collections are going to be merged before the workflow is available?

I didn't understood your question. The `update-collections` is the one who merges the collections in Firestore. If it fails in the tests lane, it doesn't continue with the merging process.

matuella:
This workflow will only work when changes are pushed to `main` branch. All collections here in this pull will be available at the same time as the workflow in the `main` branch, which, in my naive perspective, wouldn't run the sync, as it will only detect such changes, in that particular folder, after it has been available in `main`.

TLDR: shouldn't we split merging the collections after the server is available in main?

ggirotto:
Actually I believe they'll be synched, because the workflow runs on every push on main detecting the changes since the last merge commit, which includes all added collections. It'll compare the merge commit in `main` with [b15d97](https://github.com/olmps/memo/commit/b15d976f7b7830cee7cac48623ab4dbf096b910c), which is the last commit, and in this difference these added collections will be considered.

matuella (approved):
Then I believe we should test it and see if it working as expected, otherwise split the `server` merge into the actual changes and the collections to-be-added.

I'll aprove this nonetheless, but please keep this is mind.

PR #274, opened 2021-12-09
Adds missing `setDoc` and `deleteRecursively` tests

ggirotto:


ggirotto:
@matuella I'm not sure why you left these tests behind, were you thinking in some more complex tests? Because in my observation they're almost identical with the other unit tests we already implemented in the same file.

matuella (approved):
> @matuella I'm not sure why you left these tests behind, were you thinking in some more complex tests? Because in my observation they're almost identical with the other unit tests we already implemented in the same file.

Nope! It was just because I literally forgot and didn't wanted to go back (stopping what I was doing) to it. Looks good!

PR #277, opened 2021-12-09
Fixes assertions awaits

ggirotto:
Fixes missing `assert.reject` awaits, which was causing false-positive when running tests scripts

### DIFF
--- a/firebase/test/core/data/schemas/memo.spec.ts
+++ b/firebase/test/core/data/schemas/memo.spec.ts
@@ -1,6 +1,6 @@
+import * as assert from "assert";
 import Ajv2020 from "ajv/dist/2020";
 import { SchemaValidator } from "#data/schemas/schema-validator";
-import { doesNotThrow, throws } from "assert";
 import SerializationError from "#faults/errors/serialization-error";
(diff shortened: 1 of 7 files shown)


firebase/test/core/data/schemas/memo.spec.ts
```diff
@@ -1,6 +1,6 @@
+import * as assert from "assert";
 import Ajv2020 from "ajv/dist/2020";
 import { SchemaValidator } from "#data/schemas/schema-validator";
-import { doesNotThrow, throws } from "assert";
```
ggirotto, line 3: Updated to the pattern that we are using: `import * as assert from "assert";`

firebase/test/entity-tester.ts
```diff
@@ -1,6 +1,6 @@
-import Ajv2020 from "ajv/dist/2020";
 import * as Joi from "joi";
-import { doesNotThrow, throws } from "assert";
+import * as assert from "assert";
+import Ajv2020 from "ajv/dist/2020";
```
ggirotto, line 3: Updated tha pattern

ggirotto:
> these aren't mocks per-se, I would say fakes are a better naming for such variables.

Agreed. Renamed.

PR #279, opened 2021-12-09
Adds `SyncCollectionUseCase` tests

ggirotto:


firebase/test/core/domain/use-cases/sync-collections.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+    const memosRepo = (memosRepoStub = createSinonStub(MemosRepository, sandbox));
+    const gitRepo = (gitRepoStub = createSinonStub(GitRepository, sandbox));
+
+    syncCollectionsUseCase = new SyncCollectionsUseCase(localCollections, storedCollections, memosRepo, gitRepo);
+  });
+
+  beforeEach(() => {
+    validateLocalCollectionStub = sandbox.stub(collection, "validateLocalCollection");
+    validateStoredCollectionStub = sandbox.stub(collection, "validateStoredCollection");
+    validateMemoStub = sandbox.stub(memo, "validateMemo");
```
ggirotto, line 39: These stubs are typed. Typing anything else than the correct method name will break when trying to compile

matuella (commented):
Overall, two things:

I think we should be more incisive in tests on how memos behave, as they are a fundamental part of how this use-case works. As tests are, there is a single test on "Memos Operations". Will all memos be correctly updated when there are some that were removed? Will all of them be removed when a collection is as well? Will all of them be added when a new collection is added? Shouldn't them be tested alongside its collection updates as well? What if you isolate these "tests" (of memos + collections) and they don't work as intended when "grouped" together? I feel like there are a couple of questions that can be definitely answered in these tests, but they aren't being addressed.

And the second is the "Failures" group tests, which looks compact and readable, although the "should ..." description should be a little less technical, like `should fail when GitRepository.lastCommitHash rejects` could be `should throw when failing to retrieve the last commit hash`.

matuella (approved):
Looking good! Thanks for the effort!

PR #281, opened 2021-12-09
General Fixes to SyncCollectionsUseCase

ggirotto:


firebase/src/core/data/gateways/env-gateway.ts
```diff
@@ -20,7 +20,7 @@ export class EnvGateway {
 
   /** Returns `true` if running in local environment, i.e, under Firebase Emulators. */
   get isLocalDevelopment(): boolean {
-    return process.env["FUNCTIONS_EMULATOR"] === "true";
+    return process.env["FUNCTIONS_EMULATOR"] === "true" || process.env["ENVIRONMENT"] === "DEV";
```
ggirotto, line 23: When debugging scripts, even if the Firebase Emulator is running, the env `FUNCTIONS_EMULATOR` is false because we need to run the script in a different process (i.e, a different terminal window). To solve this issue, I suggest to add another env variable `ENVIRONMENT`, which must be set when running scripts and other executions that don't belong to firebase emulators itself

firebase/src/core/domain/models/memo.ts
```diff
@@ -7,7 +7,8 @@ interface MemoContent {
     bold?: boolean;
     italic?: boolean;
     underline?: boolean;
-    codeBlock?: boolean;
+    // Quill uses `kebab-case` format for attributes
+    "code-block"?: boolean;
```
ggirotto, line 11: Updated to follow the same naming as `quill`

ggirotto: We could do as we are doing with the collections (Local & Stored), but all the effort to separate the `Memo` entity into `LocalMemo` and `StoredMemo` doesn't seem to worth it IMO.

firebase/src/core/domain/use-cases/collections/sync-collections.ts
```diff
@@ -83,11 +83,19 @@ export class SyncCollectionsUseCase {
...
+        description: localCollection.description,
+        tags: localCollection.tags,
+        category: localCollection.category,
+        locale: localCollection.locale,
+        contributors: localCollection.contributors,
+        resources: localCollection.resources,
         memosAmount: memosOrder.length,
         memosOrder: memosOrder,
       };
+
```
ggirotto, line 98: Had to update removing the spread operator because using like `...localCollection` adds `memos` property to the new object, which shouldn't have the memos.

firebase/src/core/utils/validate.ts
```diff
@@ -11,7 +11,7 @@ export const defaultMaxStringLength = 255;
  * @param allowUnknown Allows object to contain unknown keys which are ignored.
  * @param convert Attempts to cast values to the required types.
  */
-export function validate(schema: Joi.Schema, data: unknown, allowUnknown = true, convert = false): void {
+export function validate(schema: Joi.Schema, data: unknown, allowUnknown = false, convert = false): void {
```
ggirotto, line 14: Here I strongly suggest to be less permissive by default. I had some false-positives in the tests when turning this to `false` that made me think that we should be explicit when we want to allow unknown properties, and not the opposite.

firebase/test/core/domain/use-cases/sync-collections.spec.ts
```diff
@@ -8,7 +8,7 @@ import { MemosRepository } from "#data/repositories/memos-repository";
 import { GitRepository } from "#data/repositories/git-repository";
```
ggirotto, line 8: Tests needed to be updated as a consequence of this update

firebase/src/core/domain/models/memo.ts
ggirotto: Exactly, that's why seems to be not worth it to me. The only difference would be this specific `code-block` property.

matuella: Why the `code-block` would be different from a local to a stored memo?

ggirotto: It's just the naming convention. We would have:

```typescript
class LocalMemo {
  "code-block": boolean
}

class StoredMemo {
  codeBlock: boolean
}
```

firebase/src/core/data/gateways/env-gateway.ts
ggirotto: You're right, this `ENVIRONMENT` doesn't make sense. I'm thinking in another possibility to "protect" the scripts from being run in production when they should be run locally.

firebase/src/core/data/gateways/env-gateway.ts
```diff
@@ -20,7 +20,10 @@ export class EnvGateway {
 
   /** Returns `true` if running in local environment, i.e, under Firebase Emulators. */
   get isLocalDevelopment(): boolean {
-    return process.env["FUNCTIONS_EMULATOR"] === "true";
+    const isProduction = process.env["NODE_ENV"] === "production";
+    const isCi = process.env["CI"] === "true";
+
+    return !isProduction && !isCi;
```
ggirotto, line 26: Updated to use `NODE_ENV` (set when running in Google Cloud environment) and `CI` (set when running in any CI machine) instead `FUNCTIONS_EMULATOR` env variable. This way we can infer if we are running in Google Cloud or in a CI machine without "injecting" extra environment variables in the scripts

ggirotto:
@matuella Updated

firebase/src/core/domain/models/memo.ts
matuella: Oh sure, we can assume all of these as the raw quill metadata using their raw representation

matuella (approved):

PR #283, opened 2021-12-09
Adds firebase `release` workflow

ggirotto:


firebase/package.json
```diff
@@ -13,7 +13,7 @@
     "lint:core": "eslint 'src/core/**/*.ts'",
     "lint:scripts": "eslint 'src/scripts/**/*.ts'",
     "lint:test": "eslint 'test/**/*.ts'",
-    "deploy": "firebase deploy",
+    "deploy": "firebase deploy --project memo-312313",
     "deploy:firestore": "firebase deploy --only firestore",
     "deploy:functions": "firebase deploy --only functions",
```
matuella, line 18: If deploying wilkl only occur through GH actions, and we are only running `npm deploy`, I think we can remove both scripts below

matuella (commented):
Looking good!!

matuella (approved):

